PR #6, opened 2022-02-07
FI-1407: Use oauth_credentials

Jammjammjamm:
Use `oauth_credentials` so that access tokens can automatically refresh.

### DIFF
--- a/lib/us_core/generated/us_core_test_suite.rb
+++ b/lib/us_core/generated/us_core_test_suite.rb
@@ -61,13 +61,15 @@ module USCore
       title: 'FHIR Endpoint',
       description: 'URL of the FHIR endpoint',
       default: 'https://inferno.healthit.gov/reference-server/r4'
-    input :bearer_token,
-      title: 'Bearer Token',
-      default: 'SAMPLE_TOKEN'
+    input :smart_credentials,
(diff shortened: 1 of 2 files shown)


lib/us_core/generated/us_core_test_suite.rb
```diff
@@ -61,13 +61,15 @@ def self.metadata
...
       description: 'URL of the FHIR endpoint',
       default: 'https://inferno.healthit.gov/reference-server/r4'
-    input :bearer_token,
-      title: 'Bearer Token',
-      default: 'SAMPLE_TOKEN'
+    input :smart_credentials,
+      title: 'OAuth Credentials',
+      type: :oauth_credentials,
+      optional: true,
+      default: JSON.generate(access_token: 'SAMPLE_TOKEN')
```
arscan, line 68: Minor point, but is this this 'correct' way of doing this, or would you usually want to instantiate a credential-typed object and to-json it?

arscan (approved):

lib/us_core/generated/us_core_test_suite.rb
Jammjammjamm: No, that would be better.

PR #9, opened 2022-02-15
FI-1397 Add Multiple-or test to us-core test kit

yunwwang:
This PR migrate multiple-or tests from Inferno Program to US Core test kit

### DIFF
--- a/lib/us_core/generated/goal/metadata.yml
+++ b/lib/us_core/generated/goal/metadata.yml
@@ -144,7 +144,7 @@
   :path: target.measure
 - :type: Quantity
   :strength: example
-  :system: 
+  :system:
   :path: target.detail
 - :type: CodeableConcept
   :strength: example
--- a/lib/us_core/generated/goal/metadata.yml
+++ b/lib/us_core/generated/goal/metadata.yml
@@ -144,7 +144,7 @@
   :path: target.measure
 - :type: Quantity
   :strength: example
-  :system:
+  :system: 
   :path: target.detail
 - :type: CodeableConcept
   :strength: example
--- a/lib/us_core/search_test.rb
+++ b/lib/us_core/search_test.rb
@@ -315,7 +315,7 @@ module USCore
     def perform_multiple_or_search_test
       resolved_one = false
 
-      all_search_params.flat_map do |patient_id, params_list|
+      all_search_params.each do |patient_id, params_list|
         search_params = params_list.first
         existing_values = {}
         missing_values = {}
--- a/lib/us_core/search_test.rb
+++ b/lib/us_core/search_test.rb
@@ -89,7 +89,7 @@ module USCore
 
       skip_if resources_returned.empty?, no_resources_skip_message
 
-      perform_multiple_or_search_test
+      perform_multiple_or_search_test if multiple_or_search_params.present?
     end
 
     def perform_search(params, patient_id)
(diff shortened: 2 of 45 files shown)


lib/us_core/generated/goal/metadata.yml
```diff
@@ -144,7 +144,7 @@
   :path: target.measure
 - :type: Quantity
   :strength: example
-  :system:
+  :system:
```
Jammjammjamm, line 147: I don't know where these trailing spaces are coming from, but it doesn't seem like they should be here.

lib/us_core/search_test.rb
```diff
@@ -298,12 +301,54 @@ def status_search_param_name
...
       return [] if definition.blank?
 
       definition[:multiple_or] == 'SHALL' ? [definition[:values].join(',')] : [definition[:values]]
     end
 
+
+    def perform_multiple_or_search_test
+      resolved_one = false
+
+      all_search_params.flat_map do |patient_id, params_list|
```
Jammjammjamm, line 318: Shouldn't this just be an `each`? It doesn't seem like anything is being done with the return value.

lib/us_core/search_test.rb
```diff
@@ -298,12 +301,54 @@ def status_search_param_name
...
+
+        resources_returned =
+          fetch_all_bundled_resources
+            .select { |resource| resource.resourceType == resource_type }
+
+        multiple_or_search_params.each do |param_name|
+          missing_values[param_name.to_sym] = existing_values[param_name.to_sym] - resources_returned.map(&param_name.to_sym)
+        end
+
+        missing_value_message = missing_values.reject { |_k, v| v.empty? }.map { |k, v| "#{v.join(',')} values from #{k}" }.join(' and ')
```
Jammjammjamm, line 344: Could you use meaningful variable names rather than `k` and `v`?

lib/us_core/generated/goal/metadata.yml
yunwwang: This is generated automatically. Don't know why the nil element is saved with a trailing blank space. Manually removed for now.

lib/us_core/search_test.rb
yunwwang: The code was copied from run_search_test() at line 85 https://github.com/inferno-framework/us-core-test-kit/blob/ff3902817d3230eda47c5b59a6778cd078e296ac/lib/us_core/search_test.rb#L85

Jammjammjamm: It's being assigned to something there. It isn't here.

lib/us_core/search_test.rb
```diff
@@ -298,12 +301,57 @@ def status_search_param_name
...
+      resolved_one = false
+
+      all_search_params.each do |patient_id, params_list|
+        search_params = params_list.first
+        existing_values = {}
+        missing_values = {}
+
+        multiple_or_search_params.each do |param_name|
+          search_value = default_search_values(param_name.to_sym)
+          search_params = search_params.merge("#{param_name}" => search_value)
```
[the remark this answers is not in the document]
Jammjammjamm: The same happens in a bunch of the observation tests.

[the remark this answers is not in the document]
Jammjammjamm: Adding the `next` will prevent this bug from happening, but I'm concerned that we're just covering it up and this could hide future bugs. Why was this nil error happening with the multiple or tests and not with any of the others?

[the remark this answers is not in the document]
yunwwang: Other searches loops through params_list array so if the params_list is empty then loop ends automatically. multiple_or test does not loop through all search_params hash in the params_list array instead it takes the first search_params hash and build its own search paramasters.

Jammjammjamm (approved):

PR #17, opened 2022-03-16
Generate US Core v4.0.0 tests

yunwwang:
Generator tests for US Core v4.0.0 based on StructureDefinitions

Changes in this PR include:
* Update US Core Generator to adopt US Core StructureDefinition Changes
* Remove most MustSupport.element.path in metadata.yml (Except for Reference) to reduce file size
* Update reference_test to handle type-must-support target_profile check
* Update search_test to handle search parameters with multiple path


Jammjammjamm (commented):
Everything seems to be working and is good overall. The generator needs to be rerun as there is a change which isn't in the generated files.

chris-at-work (commented):
There's a refactor of that map/each I want to do.  I'll push a commit to this branch.

Jammjammjamm (approved):

PR #18, opened 2022-03-23
Fix full_path parsing on expressions with two types

chris-at-work:
Fix reported bug

> The method full_path does not handle search parameter having multiple types correctly. Ex US Core condition-onset-date:
> "expression": "Condition.onset.as(dateTime)|Condition.onset.as(Period)"
> and full_path returns:
> ["Condition.onsetDateTime", "Condition.onsetDateTime"]


yunwwang (approved):
It works!

PR #19, opened 2022-03-25
FI-1502 Handles Must Support Choice

yunwwang:
US Core use narrative section to fine turn some MustSupport elements, such as:

Although both are marked as must support, servers are not required to support both Encounter.location.location and Encounter.serviceProvider, but they SHALL support at least one of these elements.

This PR added "choices" to metadata to indicate which elements are in the choice pair. For example, in the encounter's metadata.yml

```
:must_supports:
  :choices:
  - :paths:
    - location.location
    - serviceProvider
```

must_support_test remove a miss element is it is one of the choice element and the other choice element is not missing.

This PR shall be reviewed and merged after PR #17

### DIFF
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
@@ -271,11 +271,12 @@ module USCoreTestKit
         remove_vital_sign_component
         remove_blood_pressure_value
         remove_observation_data_absent_reason
-        remove_document_reference_attachment_data_url
+        #remove_document_reference_attachment_data_url
+        add_must_support_choices
 
         case profile.version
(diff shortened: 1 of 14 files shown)


lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
```diff
@@ -271,11 +271,12 @@ def handle_special_cases
         remove_vital_sign_component
         remove_blood_pressure_value
         remove_observation_data_absent_reason
-        remove_document_reference_attachment_data_url
+        #remove_document_reference_attachment_data_url
```
Jammjammjamm, line 274: We can get rid of these now, right?

Jammjammjamm (approved):

PR #25, opened 2022-03-31
FI-1514: Fix escaped string interpolation

Jammjammjamm:
Fix an incorrectly escaped string interpolation which prevents absolute urls from matching search parameters. Raised in https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/63


lib/us_core_test_kit/search_test.rb
```diff
@@ -682,7 +682,7 @@ def check_resource_against_params(resource, params)
               # references can also be URL's, so we made need to resolve those url's
```
arscan, line 682:
```suggestion
              # references can also be URLs, so we may need to resolve those URLs
```

arscan (approved):

yunwwang (approved):

PR #34, opened 2022-04-20
FI-1539 Omit medication resource validation if no medication resources provided.

arscan:
Test 13.08 in the v3.1.1 US Core Test kit should not 'skip' systems that do not provide any medication resources, because they are optional to implement (you can provide `MedicationRequest.medicationCodeableConcept` instead of a reference to a Medication resource per US Core).  This also applies to US Core 4.0.0.

To test, run against the Reference server against only patient 355.  Previously, the test would skip against this resource, because it provides no example of Medication references in MedicationRequest.  Now, it will omit.  We didn't notice this because we typically test against both patients 85 and 355, and patient 85 provides an example of this.

This was reported against the g10 test kit here: https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/101

### DIFF
--- a/lib/us_core_test_kit/validation_test.rb
+++ b/lib/us_core_test_kit/validation_test.rb
@@ -3,9 +3,15 @@ module USCoreTestKit
     DAR_CODE_SYSTEM_URL = 'http://terminology.hl7.org/CodeSystem/data-absent-reason'.freeze
     DAR_EXTENSION_URL = 'http://hl7.org/fhir/StructureDefinition/data-absent-reason'.freeze
 
-    def perform_validation_test(resources, profile_url)
-      skip_if resources.blank?,
-              "No #{resource_type} resources conforming to the #{profile_url} profile were returned."
+    def perform_validation_test(resources,
+                                profile_url,
+                                must_demonstrate_resource_type: true)
+
+      skip_if must_demonstrate_resource_type && resources.blank?,
+              "No #{resource_type} resources conforming to the #{profile_url} profile were returned"
+
+      omit_if resources.blank?,
+              "No #{resource_type} resources provided so the #{profile_url} profile does not apply"
 
       resources.each do |resource|
         resource_is_valid?(resource: resource, profile_url: profile_url)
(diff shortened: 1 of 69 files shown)


lib/us_core_test_kit/validation_test.rb
```diff
@@ -3,9 +3,15 @@ module ValidationTest
     DAR_CODE_SYSTEM_URL = 'http://terminology.hl7.org/CodeSystem/data-absent-reason'.freeze
     DAR_EXTENSION_URL = 'http://hl7.org/fhir/StructureDefinition/data-absent-reason'.freeze
 
-    def perform_validation_test(resources, profile_url)
-      skip_if resources.blank?,
-              "No #{resource_type} resources conforming to the #{profile_url} profile were returned."
+    def perform_validation_test(resources,
+                                profile_url,
+                                must_demonstrate_resource_type: true)
```
yunwwang, line 8: I don't like this parameter name. I prefers more simple name such as "skip_if_empty". Just my personal preference.

arscan: I named it this because it seemed better to me for the test to deal with implementation details of how to enforce this requirement.  That way the code that calls it doesn't have to remember what "skip" vs "omit" means. It just needs to know whether or not at least one resource needs to be provided in order for the test to pass.  If @Jammjammjamm agrees with @yunwwang though I'll change it (2 vs 1) :)

Jammjammjamm: I agree with Yunwei.

yunwwang: It is not clear, just from the name `must_demonstrate_resource_type`, what this parameter is.
how about `resources_must_exist`?

Jammjammjamm (commented):
Please put the ticket in the PR title.

lib/us_core_test_kit/validation_test.rb
arscan: The masses have spoken, I'll change it to `skip_if_empty`:) Stand by...

arscan: Updated to `skip_if_empty` @yunwwang @Jammjammjamm

Jammjammjamm (approved):

PR #38, opened 2022-04-29
FI-1554 clinical note attachment

yunwwang:
This PR address GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/113

The changes include:
* Update clinical_note_type_test to filter DiagnosticReport with required categories only
* Update clinical_note_type_test to not save attachment without url
* Add unit test


nathanloyer (approved):
Looks like this should resolve the issue I reported

nathanloyer:
Just deployed and tested this version and confirmed it gets our tests passing. 

@yunwwang, just want to make sure this will make it in the may release for the g10 kit. We're going to need this for our certification runs later this month.

arscan (approved):

PR #41, opened 2022-05-18
Fi-1505 date search precision

yunwwang:
US Core v4.0.0 requires that

When searching using the date type searchparameter [(how to search by date)](http://hl7.org/fhir/R4/search.html#date):
The client SHALL provide values precise to the day for elements of datatype date and to the second + time offset for elements of datatype dateTime.
The server SHALL support values precise to the day for elements of datatype date and to the second + time offset for elements of datatype dateTime.

This PR adds logic that

IF the value is year (YYYY) or month (YYYY-mm) precision, the search value is converted to greater than range search with precision of second+offset
IF the value is date (YYYY-mm-dd) AND the resource is NOT Goal, the search value is converted to greater than range search with precision of second+offset
OTHERWISE the value has the correct precision and can be used directly.


Jack-Fraser16:
Code looks good, but I guess I'm a little fuzzy on the logic here. For datatype date, why are you adding precision to second + time offset? I read the guidance as saying datatype date only needs be precise to the day.

lib/us_core_test_kit/search_test.rb
```diff
@@ -568,7 +568,24 @@ def search_param_value(name, patient_id, include_system: false)
...
+               params_with_comparators&.include?(name)
+              # convert date search to greath-than comparator search with correct precision
+              # For all date search parameters:
+              #   Patient.birthDate does not mandate comparators so cannot be converted
+              #   Goal.target-date has day precision
+              #   All others have second + time offset precision
+              if /^\d{4}$/.match?(element) || # YYYY
+                 /^\d{4}-\d{2}$/.match?(element) || # YYYY-MM
+                (/^\d{4}-\d{2}-\d{2}$/.match?(element) && resource_type != "Goal") # YYYY-MM-DD AND Resource is NOT Goal
+                "gt#{(DateTime.xmlschema(element)-1).xmlschema}"
```
Jack-Fraser16, line 582:
```suggestion
                "gt#{(DateTime.xmlschema(element)-1).xmlschema}"
```
Could you also explain the logic behind subtracting 1? I'm assuming it has to do with subtracting by some amount so that the greater than comparator returns what you'd expect, but I want to confirm.

yunwwang: The logic is to convert year/month/day to range compare to cover the correct date time range to ensure that search has at least one result

2020 => gt2019-12-31T00:000:00 # so the resource with date value '2020' is in the result

2020-04 => gt2020-03-31T00:00:00

2020-04-12 => gt2020-04-11T00:00:00

yunwwang: > Code looks good, but I guess I'm a little fuzzy on the logic here. For datatype date, why are you adding precision to second + time offset? I read the guidance as saying datatype date only needs be precise to the day.

That is a good point. I will fix that.

yunwwang: We use time+offset in comparator search for both the date and date-time precision. After some investigation, I think our solution is OK.
1) time+offset is a valid presentation of date search parameter, which is defined as "The date parameter format is yyyy-mm-ddThh:mm:ss[Z|(+|-)hh:mm] (the standard XML format)."
2) comparator + time-offset presents a valid and meaning range search. For example: gt2022-05-14T00:00:00+00:00 is a valid range without any ambiguity no matter for date or datetime target value.

spec/us_core/search_test_spec.rb
```diff
@@ -392,4 +392,249 @@ def scratch_resources
...
+        stub_request(:get, "#{url}/Goal?patient=#{patient_id}&target-date=le2020-03-05T13:01:01-04:00")
+          .to_return(status: 200, body: bundle_datetime.to_json)
+
+        result = run(test_class, patient_ids: patient_id, url: url)
+        expect(request).to have_been_made.once
+        expect(result.result).to eq('pass')
+      end
+    end
+
+    context 'with datetime precision' do
```
Jack-Fraser16, line 522:
```suggestion
    context 'with dateTime precision' do
```

Jack-Fraser16 (approved):
Looks good. My only comment is that stubbing the comparator search requests only tests whether the correct request was made - if some error causes no comparator requests to be made, that error wouldn't be caught. This seems to be an issue throughout the search unit tests.

PR #43, opened 2022-06-01
FI-1557 Fix Patient's previous name Must Support test

yunwwang:
This PR updates testing logic for patient's previous name MustSupport test
* US Core 4.0 does not list patient's previous name as a MustSupport in StructureDefinition
* US Core 4.0 mentioned patient's previous as "Additional USCDI requirement"
* US Core 4.0 states that "Previous name is represented by providing an end date in the Patient.name.period element for a previous name."
* ONC CCG clarifies that "Either the US Core IG Profile “StructureDefinition-us-core-patient” element “name.period” or “name.use” is required for testing and certification in the ONC Certification Program to meet the USCDI requirement to support the “Patient Demographics” data class: “Previous Name” data element"

This PR includes
* Adds both name.period.end and name.use:old as MustSupport elements
* Adds name.period.end and name.use as MustSupport choices

### DIFF
--- a/lib/us_core_test_kit/generated/v4.0.0/metadata.yml
+++ b/lib/us_core_test_kit/generated/v4.0.0/metadata.yml
@@ -12119,13 +12119,10 @@
     :extensions:
     - :id: Patient.extension:race
       :url: http://hl7.org/fhir/us/core/StructureDefinition/us-core-race
-      :uscdi_only: true
     - :id: Patient.extension:ethnicity
       :url: http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity
-      :uscdi_only: true
     - :id: Patient.extension:birthsex
       :url: http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex
-      :uscdi_only: true
     :slices: []
     :elements:
     - :path: identifier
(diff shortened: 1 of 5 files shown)


lib/us_core_test_kit/generated/v4.0.0/metadata.yml
```diff
@@ -12143,10 +12143,14 @@
     - :path: address.postalCode
     - :path: address.period
     - :path: communication.language
-    - :path: telecom
```
Jammjammjamm, line 12146: Do you know why this changed?

yunwwang: That was my understanding. 
In US Core 3.1.1 telecom, telecom.system, telecom.value, telecom.use are MS. 
In US Core 4.0.0 telecom is not MS but his three subelements are still MS.
In US Core 4.0.0, the "Additional USCDI requirement" indicate "contact detail (e.g. a telephone number or an email address)" as one of the additional USCDI element.

We can
1) add MS back to telecom, and communication, or
2) leave as is because if server does not populate telecom element, then MS test fails on telecom.system, telecom.name, and telecom.use. 

So Inferno can catch if system does not support USCDI patient's contact detail no matter telecom is a MS or not.

Jammjammjamm: I think it's fine. I just missed the code change that led to this, and I see it now.

Jammjammjamm (approved):

PR #44, opened 2022-06-02
FI-1585 Separate USCDI requirements from USCore requirements

yunwwang:
USCore v4.0 starts separation of USCDI requirements into "Additional USCDI requirement" narrative section. 

This PR changes
* Add uscdi_only tag to those USCDI requirements 
* Add logic in must_support_test to load or not load USCDI tests based on configuration
* Update Must Support test description to concatenate MustSupport choices with "or"

This PR depends on #43

### DIFF
--- a/lib/us_core_test_kit/generator/must_support_test_generator.rb
+++ b/lib/us_core_test_kit/generator/must_support_test_generator.rb
@@ -75,25 +75,21 @@ module USCoreTestKit
         build_must_support_list_string(true)
       end
 
-      def nil_to_false(element)
-        element.nil? ? false : element
-      end
-
       def build_must_support_list_string(uscdi_only)
         slice_names = group_metadata.must_supports[:slices]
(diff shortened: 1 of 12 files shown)


lib/us_core_test_kit/generator/must_support_test_generator.rb
```diff
@@ -68,9 +68,36 @@ def resource_collection_string
...
+        build_must_support_list_string(true)
+      end
+
+      def nil_to_false(element)
+        element.nil? ? false : element
+      end
+
+      def build_must_support_list_string(uscdi_only)
+        slice_names = group_metadata.must_supports[:slices]
+          .select { |slice| nil_to_false(slice[:uscdi_only]) == uscdi_only }
```
Jammjammjamm, line 84: I think you can get rid of `nil_to_false` using `slice[:uscdi_only].presence == uscdi_only.presence`.

```
[1] pry(main)> nil.presence
=> nil
[2] pry(main)> false.presence
=> nil
[3] pry(main)> true.presence
=> true
```

PR #46, opened 2022-06-14
FI-1595 Remove resource type check

yunwwang:
This PR fixes GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/141

The changes are:

* Remove validation of resource types in Bundle search result.


Jack-Fraser16 (approved):
Looks good to me. Do you think it needs an explicit unit test? Also, do you regenerate tests as part of this PR, or does that the responsibility of whoever uses this kit/gem?

PR #47, opened 2022-06-15
FI-1599 Skip array item with DAR extension

yunwwang:
This PR fixes part of GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/142

The changes includes
* Add parameter `include_dar: false` to method `find_a_value_at`. The method skips element with DAR extension when  `include_dar == false`
* Add unit test to cover this use case

Testing
* A unit test was created to cover the use case. The simplest testing is observing all unit tests pass
* To run integration test:
** Create a patient with name having DAR as the first item. Example
```
"name": [
      {
        "extension": [
          {
            "url": "http://hl7.org/fhir/StructureDefinition/data-absent-reason",
            "valueCode": "unknown"
          }
        ]
      },
      {
        "use": "official",
        "text": "Johnny LastName",
        "family": "LastName",
        "given": [
          "Johnny"
        ]
      }
    ],
```
** Run the Patient Tests (Test Group 2) and verify the search by name test (Test 2.03) passes

### DIFF
--- a/lib/us_core_test_kit/fhir_resource_navigation.rb
+++ b/lib/us_core_test_kit/fhir_resource_navigation.rb
@@ -22,10 +22,11 @@ module USCoreTestKit
       elements = Array.wrap(element)
 
       if path.empty?
-        elements.reject! do |el|
-          el.respond_to?(:extension) &&
-          el.extension.any? { |ext| ext.url == DAR_EXTENSION_URL}
-        end unless include_dar
+        unless include_dar
+          elements.reject! do |el|
+            el.respond_to?(:extension) && el.extension.any? { |ext| ext.url == DAR_EXTENSION_URL}
+          end
+        end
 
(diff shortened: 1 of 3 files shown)


lib/us_core_test_kit/fhir_resource_navigation.rb
```diff
@@ -14,12 +16,17 @@ def resolve_path(elements, path)
...
+    def find_a_value_at(element, path, include_dar: false)
       return nil if element.nil?
 
       elements = Array.wrap(element)
 
       if path.empty?
+        elements.reject! do |el|
+          el.respond_to?(:extension) &&
+          el.extension.any? { |ext| ext.url == DAR_EXTENSION_URL}
+        end unless include_dar
```
Jammjammjamm, line 28: `if/unless` should only go at the end of single-line statements.

Jammjammjamm (approved):

PR #51, opened 2022-06-30
FI-1607 Filter Devices in post_search and reference_with_type_search

yunwwang:
This PR fixes the issue raised in GitHub https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/149

Inferno allows tester to specify Device Type Code filter in the input dialog. If the filter values are provided, Inferno filters the Devices returned in GET search. Inferno does not filter the Device resources returned in POST search nor in the Reference Type Search.

Changes in this PR includes:
* Call filter_devices in perform_post_search()
* Call filter_devices in perform_reference_with_type_search
* Update message of no_resource_found to indicate Devices may be filtered
* Add Unit Tests for both the logic and updated skip message


lib/us_core_test_kit/search_test.rb
```diff
@@ -470,8 +474,13 @@ def empty_search_params_message(empty_search_params)
     end
 
     def no_resources_skip_message(resource_type = self.resource_type)
-      "No #{resource_type} resources appear to be available. " \
-      "Please use patients with more information"
+      msg = "No #{resource_type} resources appear to be available "
```
Jammjammjamm, line 477:
```suggestion
      msg = "No #{resource_type} resources appear to be available"
```

lib/us_core_test_kit/search_test.rb
```diff
@@ -240,10 +243,11 @@ def perform_reference_with_type_search(params, resource_count)
       new_search_params = params.merge('patient' => "Patient/#{params['patient']}")
       search_and_check_response(new_search_params)
 
-      new_resource_count =
-        fetch_all_bundled_resources
-          .select { |resource| resource.resourceType == resource_type }
-          .count
+      reference_with_type_resources = fetch_all_bundled_resources.select { |resource| resource.resourceType == resource_type }
+
+      filter_devices(reference_with_type_resources) if resource_type == 'Device'
```
Jack-Fraser16, line 248: So we added filter here (which makes sense) - why isn't it added into the `perform_search_with_system` method?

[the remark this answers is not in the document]
yunwwang: perform_search_with_system() only checks if the returned resources are not empty. In such case, we don't need to worry about further filtering the resources.

PR #53, opened 2022-07-07
Release v0.2.4

Jack-Fraser16:
Main purpose is to patch Validation Message Filtering. 
To test - open G10 and point US Core gem to this PR.


yunwwang (approved):

arscan (approved):
I also looked at the other 2 filters that are listed in there but they appear to not be set to start matching at the start of the string (just must end at the end), so the update to inferno-core shouldn't affect them.

PR #55, opened 2022-07-14
FI-1634: Multi version validation

Jammjammjamm:
This branch updates the validation tests to include the profile version in their validation calls. This lets us get rid of the separate validators for US Core 3 & 4 and just use a single validator.

This will address https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/165.


yunwwang (commented):
When I run "bundle exec rake us_core:generate" I got an error

Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/yunweiw/GitHub/Inferno/us-core-test-kit/lib/us_core_test_kit/igs/v311/package.tgz

arscan (commented):
I confirmed that the tests now send along the profile version, and can't think of any other places where this may need to be changed in this kit.  We'd have to make sure that all the Bulk Data tests in g10 get updated to include version.

Jammjammjamm:
> We'd have to make sure that all the Bulk Data tests in g10 get updated to include version.

Yup, that will need to be done under a separate ticket.

yunwwang (approved):

PR #57, opened 2022-07-26
include data absent reasons in checks for must support elements

nathanloyer:
https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/182


nathanloyer:
I'm still testing this to see if it resolves the issue

nathanloyer:
This got the tests passing again for our server.

image

I can add test coverage for this if yall agree this is the right thing to do

nathanloyer:
See the linked ticket for more detailed discussion on this. We've decided to make an update to our fhir server instead of merging this

PR #58, opened 2022-07-28
USCore-5

yunwwang:
This PR creates US Core v5.0.1 Test kit.

It includes
* Generate US Core v5.0.1 Test Kit from IG package
* Updates generator to support pattern slicing with required binding (FI-1624)
* Adds Must Support test for Patient's Previous Address (FI-1627)
* Supports Must Support choices between Must Support elements, extensions, slices (FI-1633)
* Blocks test generator from generating QuestionnaireResponse test group (FI-1635)
* Includes USCDI MustSupport tests by default (FI-1630)


lib/us_core_test_kit/generator/group_metadata.rb
```diff
@@ -29,14 +29,25 @@ class GroupMetadata
...
-      NON_USCDI_RESOURCES = [
-        'Encounter',
-        'Location',
-        'Organization',
-        'Practitioner',
-        'PractitionerRole',
-        'Provenance'
-      ].freeze
+      NON_USCDI_RESOURCES = {
+        'v3.1.1' => [
```
arscan, line 33: Nit pick: when reading this i thought you forgot about v4.0.0, but then see later that you consider v3.1.1 here to actually mean v3.1.1 and v4.0.0.  I wonder if you should just be a little duplicative and explicitely write out v3.1.1, v4.0.0, v5.01 here to be clear you didn't forget about v4.0.0, and that simplies logic a little bit later when calling this.

lib/us_core_test_kit/generator/ig_metadata_extractor.rb
```diff
@@ -28,8 +28,9 @@ def resources_in_capability_statement
 
       # The US Core Server Capability Statement does not list support for the
       # required vital signs profiles, so they need to be added
-      def add_vital_signs_profiles
-        if ig_resources.ig.version == "3.1.1"
+      def add_missing_supported_profiles
+        case ig_resources.ig.version
+        when '3.1.1'
```
arscan, line 33: 4.0.0 not needed here?

lib/us_core_test_kit/generator/ig_resources.rb
```diff
@@ -32,7 +32,9 @@ def value_set_by_url(url)
       end
 
       def search_param_by_resource_and_name(resource, name)
-        normalized_name = name.to_s == '_id' ? 'id' : name.to_s
+        #normalized_name = name.to_s == '_id' ? 'id' : name.to_s
```
arscan, line 35: Why leave this commented out?

Also, might be good to write a comment here to identify which search param other than _id this now applies to, as an example (I just don't know which that might be).

lib/us_core_test_kit/generator/suite_generator.rb
```diff
@@ -81,11 +81,19 @@ def capability_statement_group_id
       end
 
       def clinical_notes_guidance_file_name
-        "../../custom_groups/#{ig_metadata.ig_version}/clinical_notes_guidance_group"
+        if ig_metadata.ig_version == 'v3.1.1'
+          "../../custom_groups/#{ig_metadata.ig_version}/clinical_notes_guidance_group"
+        else
+          '../../custom_groups/v4.0.0/clinical_notes_guidance_group'
```
arscan, line 87: So v5.0.0 is writing guidance to v4.0.0 clinical notes guidance group?

lib/us_core_test_kit/generator/group_metadata.rb
yunwwang: v311 and v400 have the same set of NON_USCDI_RESOUCES while v500 is different. I will reorganize the hash and logic.

lib/us_core_test_kit/generator/ig_metadata_extractor.rb
yunwwang: Yes. This method is used to amend IG's CapabilityStatement.

For US Core v3.1.1, its CapabilityStatement does not include those FHIR vital sign profiles
For US Core v5.0.1, its CapabilityStatement missed support-profile for Encounter resource

I will add more comments in the code

lib/us_core_test_kit/generator/suite_generator.rb
yunwwang: Yes. They have the same set of required categories

lib/us_core_test_kit/generated/v4.0.0/metadata.yml
```diff
@@ -12166,11 +12166,8 @@
     - :path: name.family
     - :path: name.given
     - :path: telecom.system
-      :uscdi_only: true
```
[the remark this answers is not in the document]
yunwwang: telecom.system is marked as MS in US Core v4.0. It was a mistake.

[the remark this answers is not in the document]
Jammjammjamm: But now this change has been removed?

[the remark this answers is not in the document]
yunwwang: add uscdi_only back as discussed

Jammjammjamm:
The inconsistency in the naming of the Observation groups isn't great. Do you think there's a good way to make them consistent?

Screen Shot 2022-08-10 at 9 44 26 AM

yunwwang:
> The inconsistency in the naming of the Observation groups isn't great. Do you think there's a good way to make them consistent?
> 
> Screen Shot 2022-08-10 at 9 44 26 AM

I think I can move Observation to the front

Jammjammjamm (approved):
I think these new titles are a really nice improvement 👍

yunwwang:
The same title change is applied to US Core 4.0. I don't change US Core 3.1.1 since I don't want to cause any changes to the existing g10 test kit.

PR #59, opened 2022-07-29
FI-1657 Add must support test for datetime slicing

yunwwang:
This PR fixes GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/178

The root cause is that US Core must_support_test checks slicing of Date and String but not DateTime.

This PR affects US Core 4.0.0 Test Kit and US Core 5.0.1 Test Kit. It does NOT affect US Core 3.1.1 Test Kit.

The change include:
* Adds MustSupport test of slicing with type 'DateTime' 
* Adds Unit Test

### DIFF
--- a/spec/us_core/must_support_test_spec.rb
+++ b/spec/us_core/must_support_test_spec.rb
@@ -248,6 +248,22 @@ RSpec.describe USCoreTestKit::MustSupportTest do
 
         expect(result.result).to eq('pass')
       end
+
+      it 'skips if datetime format is not correct' do
+        observation.effectiveDateTime = "not a date time"
+        allow_any_instance_of(test_class)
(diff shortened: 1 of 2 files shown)


spec/us_core/must_support_test_spec.rb
```diff
@@ -143,55 +143,111 @@ def run(runnable, inputs = {})
...
+              {
+                system: 'http://snomed.info/sct',
+                code: '449868002'
+              }
+            ]
+          }
+        )
+      }
+
+      it 'passes if server suports all MS slices' do
```
Jammjammjamm, line 238: Should we have a test to show a failure on a bad dateTime slice?

Jammjammjamm (approved):

PR #62, opened 2022-08-11
FI-1674 Set US Core Observation Survey and US Core Questionnaire as choice

yunwwang:
This PR pre-applies US Core Ticket FI-37794 (https://jira.hl7.org/browse/FHIR-37794)

The changes is:
Set US Core Observation Survey and US Core QuestionnaireResponse as choices of target profiles for US Core Observation Survey and US Core SDOH Assessment profiles


Jammjammjamm:
Should we wait for a resolution on that jira ticket?

yunwwang:
Yes. We can. I assume we will get bunch of complains about the MS of QuestionnaireResponse

PR #77, opened 2022-09-12
Add US Core IG 5.0.1 Callout

Justin-Ramm:
# Summary

* Add note about support for the US Core Implementation Guide 5.0.1


Jammjammjamm (approved):
Good catch. Thank you for your contribution.

PR #82, opened 2022-10-12
FI-1752 Add filter_results and filter_smoking_status

yunwwang:
# Summary
The PR add filter_smoking_status method to filter Observation resources using code=72166-2.

# Testing Guidance

### DIFF
--- a/lib/us_core_test_kit/search_test.rb
+++ b/lib/us_core_test_kit/search_test.rb
@@ -106,8 +106,7 @@ module USCoreTestKit
 
       perform_comparator_searches(params, patient_id) if params_with_comparators.present?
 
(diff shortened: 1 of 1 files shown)


lib/us_core_test_kit/search_test.rb
```diff
@@ -173,6 +182,16 @@ def filter_conditions(resources)
       end
     end
 
+    def filter_smoking_status(resources)
```
Jammjammjamm, line 185: Shouldn't we make this only filter when performing a status search in US Core 5 rather than always filtering the smoking status results?

yunwwang: Do you mean category search?

Jammjammjamm: Yes.

yunwwang:
Need more investigation of this issue. PR withdrawn

PR #83, opened 2022-10-17
FI-1749: Update for ruby 3

Jammjammjamm:
Everything should still work as normal.


arscan (approved):
Ran the v3.1.1 tests and it seems to be working fine.

PR #85, opened 2022-10-24
FI-1768 Unescape search value

yunwwang:
# Summary
This PR fixes GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/314

Chang Log:
* Updates search_test to use unescaped search value when checking result against parameter
* Adds unit test

# Testing Guidance
All unit tests passed

### DIFF
--- a/lib/us_core_test_kit/search_test.rb
+++ b/lib/us_core_test_kit/search_test.rb
@@ -677,7 +677,9 @@ module USCoreTestKit
     #### RESULT CHECKING ####
 
     def check_resource_against_params(resource, params)
-      params.each do |name, search_value|
+      params.each do |name, escaped_search_value|
+        #unescape search value
+        search_value = escaped_search_value&.gsub('\\,', ',')
         paths = search_param_paths(name)
 
         match_found = false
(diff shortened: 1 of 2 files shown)


lib/us_core_test_kit/search_test.rb
```diff
@@ -677,7 +677,9 @@ def save_delayed_references(resources, containing_resource_type = self.resource_
     #### RESULT CHECKING ####
 
     def check_resource_against_params(resource, params)
-      params.each do |name, search_value|
+      params.each do |name, escaped_search_value|
+        #unescape search value
+        search_value = escaped_search_value&.gsub('\\,', ',')
```
Jammjammjamm, line 682: Is this the only thing we potentially need to unescape?

yunwwang: Yes. This is the only place the search value is actually "used"

Jammjammjamm: I really meant is `\\,` the only thing we need to unescape.

yunwwang: Yes. That is the only one. See line# 626.

PR #86, opened 2022-10-27
FI-1774: Support new validator

Jammjammjamm:
# Summary
Updates the issue filters to support changed messages in the new validator.

# Testing Guidance
In `docker-compose.yaml`, you should get the same results if you replace the image for the `validator_service` with:
- `infernocommunity/fhir-validator-service:v2.1.0`
- `infernocommunity/fhir-validator-service:v2.2.0`

### DIFF
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
@@ -232,8 +232,10 @@ module USCoreTestKit
         target_profiles = []
 
         type.source_hash['_targetProfile']&.each do |hash|
-          element = FHIR::Element.new(hash)
-          target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
+          if hash.present?
+            element = FHIR::Element.new(hash)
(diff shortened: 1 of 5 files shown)


lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
```diff
@@ -232,8 +232,10 @@ def handle_type_must_support_target_profiles(type, metadata)
         target_profiles = []
 
         type.source_hash['_targetProfile']&.each do |hash|
-          element = FHIR::Element.new(hash)
-          target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
+          if hash.present?
```
arscan, line 235: Do we know what changed to make generation suddenly not work?  Have we changed something since we last generated, and this fixes that?

Jammjammjamm: I don't know. Yunwei fixed it in #84, but I needed to regenerate the tests here, so I just copied that code over.

arscan (approved):

PR #91, opened 2022-11-04
FI-1786 search parameter definition values

yunwwang:
# Summary
This PR updates how the search definition metadata extractor get values for search parameter.

What are changed:
* Move all the value extraction methods to value_extractor class
* Remove values extracted from the base FHIR model if the search parameter is not 'status' or not multiple-or = SHALL

# Testing Guidance


lib/us_core_test_kit/generator/value_extractor.rb
```diff
@@ -1,11 +1,62 @@
 module USCoreTestKit
   class Generator
     class ValueExactor
-      attr_accessor :ig_resources, :resource
+      attr_accessor :ig_resources, :resource, :profile_elements, :profile_element
```
Jammjammjamm, line 4: I don't see where `profile_element` needs to have an accessor.

lib/us_core_test_kit/generator/search_definition_metadata_extractor.rb
```diff
@@ -114,70 +114,21 @@ def multiple_or_expectation
...
-
-        profile_elements.select do |element|
-          full_paths.include?(element.path) &&
-            element.sliceName.present? &&
-            (
-              element.patternCodeableConcept.present? ||
-              element.binding.present? && element.binding.strength == 'required'
-            )
+      def values_from_resource_metadata(paths)
+        if multiple_or_expectation == 'SHALL' || paths.include?('status')
```
Jammjammjamm, line 123: Is `status` a special case here becaues it's the only search param we iterate over to search and find resources?

lib/us_core_test_kit/generator/value_extractor.rb
yunwwang: Good catch. Forgot to remove it.

lib/us_core_test_kit/generator/search_definition_metadata_extractor.rb
yunwwang: There are two. One is the 'status' search parameters. The other is search parameters with multiple-or = 'SHALL'. Most of these are overlapped. 

There are some parameters, other than status, having multiple-or=SHALL. Like, MedicationRequest.intent. 

Immunizaiton.status is the only one status parameter which does not have multiple-or=SHALL

Jammjammjamm (approved):

PR #93, opened 2022-11-21
FI-1800 Remove values from optional patternCodeablConcept

yunwwang:
# Summary
This PR fixes GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/356

Some US Core v5.0.1 profiles have slicing using the same concept. For example: both Observation Social History and Observation SDOH Assessment have and optional slicing on ObservationCategory sdoh.

This causes a validation error because resources from difference profiles are mixed together.

This PR changes
* Remove optional patternCodeablConcept from search definition metadata

# Testing Guidance
US Core v5.0.1 test for Observation Social History and Observation SDOH Assessment shall not search with category value sdoh. The validation tests for Observation Social History and Observation SDOH Assessment shall pass

### DIFF
--- a/lib/us_core_test_kit/generated/v5.0.1/observation_sdoh_assessment/metadata.yml
+++ b/lib/us_core_test_kit/generated/v5.0.1/observation_sdoh_assessment/metadata.yml
@@ -83,7 +83,6 @@
     :comparators: {}
     :values:
     - survey
-    - sdoh
     :type: CodeableConcept
     :contains_multiple: true
     :multiple_or: MAY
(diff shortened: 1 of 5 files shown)


lib/us_core_test_kit/generated/v5.0.1/observation_sdoh_assessment/metadata.yml
```diff
@@ -83,7 +83,6 @@
     :comparators: {}
     :values:
     - survey
```
Jammjammjamm, line 85: Doesn't this have the problem that not all Observations with the `survey` category are SDOH assessments?

yunwwang: Observation SDOH Assessment and Observation Survey are basically the same except SDOH Assessment has an additional MS tag on `sdoh` category slicing. So this will not cause validation issue, at least in current v5.0.1

PR #94, opened 2022-11-29
FI-1792 Generator handle required binding slice

yunwwang:
# Summary
This is part of fix for GitHub Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/361.

This PR changes:
* Remove binding definition from generated metadata unless the binding strength is `required`. ONC (g)(10) Test Kit only validates `required` binding
* Add `required_binding_slice` if the binding strength is `required` and the binding is in a slice with min cardinality > 0. In US Core 5.0.1, there is one element has such attribute: Condition.category:us-core for US Core Condition Problems and Health Concerns Profile

# Testing Guidance
Visual Inspection of metadata

### DIFF
--- a/lib/us_core_test_kit/generator/terminology_binding_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/terminology_binding_metadata_extractor.rb
@@ -27,14 +27,14 @@ module USCoreTestKit
         end
       end
 
-      def element_has_option_binding_slice(element)
+      def element_has_optional_binding_slice(element)
         element.sliceName.present? && element.min == 0
       end
 
       def profile_elements_with_bindings
         profile_elements
           .select { |element| element.binding.present? && element.binding.strength == 'required'}
-          .reject { |element| element_has_fixed_value?(element) || element_has_option_binding_slice(element) }
+          .reject { |element| element_has_fixed_value?(element) || element_has_optional_binding_slice(element) }
       end
 
 
--- a/lib/us_core_test_kit/generator/terminology_binding_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/terminology_binding_metadata_extractor.rb
@@ -33,15 +33,23 @@ module USCoreTestKit
           .reject { |element| element_has_fixed_value? element }
       end
 
+      def is_required_binding_slice(profile_element)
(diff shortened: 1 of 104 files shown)


lib/us_core_test_kit/generator/terminology_binding_metadata_extractor.rb
```diff
@@ -27,21 +27,33 @@ def element_has_fixed_value?(element)
         end
       end
 
+      def element_has_optional_binding_slice(element)
```
Jammjammjamm, line 30: This method name should end in `?` since it returns true/false.

Jammjammjamm (commented):
Are any changes needed for MS slices (1 or 0 min cardinality) with required bindings?

yunwwang:
> Are any changes needed for MS slices (1 or 0 min cardinality) with required bindings?

No. Server still need to demonstrate that it support MS slicing regardless of the cardinality. For example, server still need to show at least one Problems and Health Concerns Condition resource with `category = sdoh`

lib/us_core_test_kit/generator/terminology_binding_metadata_extractor.rb
```diff
@@ -27,21 +27,33 @@ def element_has_fixed_value?(element)
...
-          {
+          binding = {
             type: element.type.first.code,
             strength: element.binding.strength,
             # Goal.target.detail has an unbound binding
             system: element.binding.valueSet&.split('|')&.first,
             path: element.path.gsub('[x]', '').gsub("#{resource}.", '')
           }
+
+          if element.sliceName.present? && element.min > 0
```
Jammjammjamm, line 52: Shouldn't this be using `element_has_optional_binding_slice?`?

yunwwang: `element_has_optional_binding_slice? ` is
```
if element.sliceName.present? && element.min == 0
```
So negation of `element_has_optional_binding_slice?` is not the same as the logic used here.

Jammjammjamm: 👍 I didn't notice the `==` vs `>`.

Jammjammjamm (approved):

PR #95, opened 2022-12-05
FI-1812 add provenance_validator for provenance-1 constraint

yunwwang:
# Summary
This PR fixes the US Core part of GitHub issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/363

This PR includes
* Remove provenance-1 error message from validator
* Add provenance_validator.rb to validation provenance-1 constraint
* Add unit test for provenance_validator

# Testing Guidance
1. Update reference server to US Core 5 data set.
2. Run US Core 5 test kit.
3. Verify that all Provenance tests pass

### DIFF
--- a/lib/us_core_test_kit/provenance_validator.rb
+++ b/lib/us_core_test_kit/provenance_validator.rb
@@ -1,0 +1,36 @@
+module USCoreTestKit
+  class ProvenanceValidator
+    include FHIRResourceNavigation
+
+    def self.validate(...)
+      new(...).validate
+    end
+
+    attr_reader :resource, :validation_messages
+
+    def initialize(resource)
+      @resource = resource
+      @validation_messages = []
+    end
+
+    def validate
+      # Invariant provenance-1 in US Core 5 causes validation error. See FHIR-39518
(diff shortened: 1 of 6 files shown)


lib/us_core_test_kit/provenance_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+      failed_provenance =
+        find_a_value_at(resource, 'agent') do |agent|
+          ['Practitioner', 'Device'].any? { |resource_type| agent.who.reference&.include?(resource_type) } &&
+          agent.onBehalfOf.nil?
+        end
+
+      if failed_provenance.present?
+        validation_messages << {
+          type: 'error',
+          message: "Rule provenance-1: 'onBehalfOf SHALL be present when Provenance.agent.who is a Practitioner or Device' Failed"
```
Jammjammjamm, line 29: We should include the resource id in the error message like we do here: https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/main/lib/onc_certification_g10_test_kit/terminology_binding_validator.rb#L43

PR #96, opened 2022-12-08
FI:1818: Save resolved reference with target_profiles

yunwwang:
# Summary
This PR saves the reference with the profiles it is validated with

# Testing Guidance


Jammjammjamm (commented):
This works for resources that resolve and validate successfully, but if the resource doesn't validate, it will still be tried repeatedly:

Screen Shot 2022-12-14 at 12 10 33 PM

We don't necessarily have to fix this. The current state of this PR is still a nice improvement.

Jammjammjamm (approved):

PR #102, opened 2023-02-01
FI-1890: Update validator regex

Jammjammjamm:
# Summary
Updates the validator regex for the updated validator.

# Testing Guidance
On main, the CapabilityStatement tests will fail with the updated validator. They will pass on this branch.


arscan (approved):
Glad this is a backwards-compatible regexp update!  I guess we could add something like `(|.*)?` or something to the end to make it clear why it is open-ended (so someone doesn't re-add it or something), but I think this is fine.

PR #103, opened 2023-02-01
FI-1888: Update profile support test

Jammjammjamm:
# Summary
Update the profile support test to not issue warnings based on supported profiles and to support a list of resources which must be supported in order to pass.

# Testing Guidance
Unit tests demonstrate the new functionality.


arscan (approved):
Looks good. I do wish it was a little more clear what is checked for the user -- it doesn't spell out whether or not in a given situation if they are testing to USCDI requirements or not (as what config variables are used is completely hidden from the end user).  But I think this is sufficient.

I tried it out on my machine and it works as expected.  Thanks!

arscan:
After approving, I was thinking about how in the 'ABOUT' section on really presents 2 options: USCDI and not USDCI, and yet downstream users of this test could list any set of resource.  It might make sense for the config to just be 'USCDI: true/false'.

PR #109, opened 2023-03-01
FI-1838 Add US Core v6 Patient USCDI elements

yunwwang:
# Summary
This is the first PR for US Core v6.0.0.

This PR changes:
* Temporary block errors from us-core-observation-occupation. This profile has a bug on slicing. This will be fixed in v6.0.0 final release.
* Create must_support_metadata_extractor_us_core_6
* Add US Core v6 Patient USCDI element
* Add capability_statement_group for US Core v6.0.0-ballot

## Note:
Many files are generated by test generator for us-core-6.0.0.

# Testing Guidance
* Start US Core test kits
* Select US Core v6.0.0-ballot test suite
* Choose reference server with v6.0.0 data (see the branch: https://github.com/inferno-framework/inferno-reference-server/tree/us-core-6)
* Most test in Test Group 1 and 2 should pass except:
  * 1.01 because local server does not support TLS
  * 2.11 because US Core v6.0.0-ballot Patient genderIdentity extension is not forward compatible. This extension will be reversed to the one in US Core v5.0.1 in the v6.0.0 release version.


Jammjammjamm:
`.env.development` needs to be updated with a validator for 6.0.0 ballot.

Jammjammjamm (approved):

PR #113, opened 2023-03-06
FI-1932 category first

yunwwang:
# Summary
The PR include these changes:
* Assign category search as the first search for both US Core Category profiles
* Assign category search as the first search for some US Core Obervation profiles
* Fix the Coverage.class name collision issue by using local_class as the path name
* Add Procedure.basedOn as USCDI only elements
* Update delayed resources to match US Core v6

# Testing Guidance

### DIFF
--- a/lib/us_core_test_kit/generator/group_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/group_metadata_extractor.rb
@@ -95,7 +95,7 @@ module USCoreTestKit
 
       ### BEGIN SPECIAL CASES ###
 
-      CATEGORY_FIRST_PROFILES_VERSION_ALL = [
+      ALL_VERSION_CATEGORY_FIRST_PROFILES = [
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-careplan',
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-diagnosticreport-lab',
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-diagnosticreport-note',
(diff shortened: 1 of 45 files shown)


lib/us_core_test_kit/generator/group_metadata_extractor.rb
```diff
@@ -95,20 +95,32 @@ def mark_mandatory_and_must_support_searches
...
+        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-condition-encounter-diagnosis' => [ 'v501' ],
+        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-condition-problems-health-concerns' => [ 'v501' ]
+      }
+
       def category_first_profile?
-        CATEGORY_FIRST_PROFILES.include?(profile_url)
+        CATEGORY_FIRST_PROFILES_VERSION_ALL.include?(profile_url) ||
+        (
+          CATEGORY_FIRST_PROFILES_VERSION_EXCLUDE.key?(profile_url) &&
+          !CATEGORY_FIRST_PROFILES_VERSION_EXCLUDE[profile_url].include?(reformatted_version)
```
Jammjammjamm, line 122: These two lines could be combined to just `CATEGORY_FIRST_PROFILES_VERSION_EXCLUDE[profile_url]&.include?(reformatted_version)`.

lib/us_core_test_kit/generator/group_metadata_extractor.rb
```diff
@@ -95,20 +95,32 @@ def mark_mandatory_and_must_support_searches
...
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab',
+        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-screening-assessment',
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-sdoh-assessment',
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-social-history',
-        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-survey'
+        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-survey',
+        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-simple-observation'
       ]
 
+      CATEGORY_FIRST_PROFILES_VERSION_EXCLUDE = {
```
Jammjammjamm, line 113: I find this name really confusing. The hash contains the versions being included, right? Maybe the two constants could be `ALL_VERSION_CATEGORY_FIRST_PROFILES` and `VERSION_SPECIFIC_CATEGORY_FIRST_PROFILES`, unless I'm misunderstanding them.

lib/us_core_test_kit/generator/group_metadata_extractor.rb
yunwwang: The logic is that a profile has the url but NOT having the version. 

The back ground is that US Core v5 introduced two Condition but it forgot to mandate patient + category search. This is fixed in v6 and presumably all the future versions. That makes v5 the only exception which does not patient+category as the first search. That is why the array's name is called  CATEGORY_FIRST_PROFILES_VERSION_EXCLUDE. 

If we use VERSION_SPECIFIC_CATEGORY_FIRST_PROFILES, the logic here is easier to write but we have to update that array with every new version.

I am open to do this either way: listing included versions vs listing excluded versions. Both have their own pros and cons.

lib/us_core_test_kit/generator/group_metadata_extractor.rb
yunwwang: Agree with the first one. See my previous comment for the reason to have the 2nd name. It is not good but reflect the logic to be used.

lib/us_core_test_kit/generator/group_metadata_extractor.rb
yunwwang: On my second thought, I decide to take the suggestion and changing the logic from exclusion to inclusion.

Jammjammjamm: That would be consistent with the other place where we do this kind of thing. You could also use a special value like `:all_versions` or `:always` to only have to maintain the list of specific versions for the profiles with version-specific logic.

Jammjammjamm (approved):

PR #133, opened 2023-10-06
FI-2163 Update CapabilityStatement Test 1.05 for supported profile

yunwwang:
# Summary
This PR is to enforce this US Core requirement "Declare support for the US Core Profile by including its official URL in the server’s CapabilityStatement.rest.resource.supportedProfile element" which can be found [here](https://hl7.org/fhir/us/core/STU4/conformance-expectations.html#profile-support--interaction-support) in US Core v4.0.0.

This change applies to US Core Test Suite for version 4.0.0 and up.

Current behavior for US Core Test Suite v3.1.1 is not changed.

Change log:
* Move existing profile_support_test.rb from custom_groups/capability_statement to custom_groups/v3.1.1
* Create new profile_support_test.rb in custom_groups/capability_statement, which
  * read supported profiles from CapabilityStatement.rest.resoruce.supportedProfile 
  * read us core profiles from test group
  * compare these two set of profiles
  * For US Core test, at least US Core Patient profile and one other US Core profile shall exist in CapabilityStatement
  * For (g)(10) test, all profiles for (g)(10) "required resource types" shall exist in CapabilityStatement
* Rename existing profile_support_test_spec.rb to us_core_v311_profile_support_test_spec.rb
* Create new profile_support_test_spec.rb to test the new profile_support_test

# Testing Guidance
CapabilityStatement test 1.05 for US Core versions, 3.1.1, 4.0.0, 5.0.1, and 6.1.10 shall pass


lib/us_core_test_kit/custom_groups/v3.1.1/capability_statement_group.rb
```diff
@@ -100,7 +100,7 @@ class CapabilityStatementGroup < Inferno::TestGroup
       test from: :us_core_fhir_version
       test from: :us_core_json_support
 
-      test from: :us_core_profile_support do
+      test from: :us_core_v311_profile_support do
```
Jammjammjamm, line 103: We'll need to override this test id to prevent it from changing.

Jammjammjamm (approved):

PR #139, opened 2023-11-20
FI-2316 Label Reference target profile as MustSupport is there is only one

yunwwang:
# Summary
This fixes https://oncprojectracking.healthit.gov/support/browse/FI-2316

When there is only one target profile for a MustSupport Reference, this profile shall also be treated as MustSupport. This does not affect MustSupport test but enforced reference validation test using the correct profile.

This change exculde US Core v3.1.1 which does not have any target_profile marked, and any FHIR base profiles.

# Testing Guidance

### DIFF
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
@@ -275,7 +275,6 @@ module USCoreTestKit
               element = FHIR::Element.new(hash)
               target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
             end
-            index += 1
           end
         end
 
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
@@ -270,8 +270,7 @@ module USCoreTestKit
         if type.targetProfile&.length == 1
           target_profiles << type.targetProfile.first
         else
-          index = 0
-          type.source_hash['_targetProfile']&.each do |hash|
+          type.source_hash['_targetProfile']&.each_with_index do |hash, index|
             if hash.present?
               element = FHIR::Element.new(hash)
               target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
@@ -261,7 +261,9 @@ module USCoreTestKit
       end
 
       def handle_type_must_support_target_profiles(type, metadata)
-        return if profile.version == '3.1.1'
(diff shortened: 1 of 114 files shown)


lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
```diff
@@ -254,17 +254,25 @@ def get_type_must_support_metadata(current_metadata, current_element)
...
 
-        type.source_hash['_targetProfile']&.each do |hash|
-          if hash.present?
-            element = FHIR::Element.new(hash)
-            target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
+        if type.targetProfile&.length == 1
+          target_profiles << type.targetProfile.first
+        else
+          index = 0
+          type.source_hash['_targetProfile']&.each do |hash|
```
Jammjammjamm, line 265: Ruby can [handle the index for you](https://ruby-doc.org/core-3.1.2/Enumerable.html#method-i-each_with_index): `&.each_with_index do |hash, index|`

lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
```diff
@@ -254,17 +254,26 @@ def get_type_must_support_metadata(current_metadata, current_element)
...
-            target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
+        if type.targetProfile&.length == 1
+          target_profiles << type.targetProfile.first
+        else
+          type.source_hash['_targetProfile']&.each_with_index do |hash, index|
+            if hash.present?
+              element = FHIR::Element.new(hash)
+              target_profiles << type.targetProfile[index] if type_must_support_extension?(element.extension)
+            end
+            index += 1
```
Jammjammjamm, line 271: This one needs to go, too.

yunwwang: Oh, got it. So the index increase automatically.

lib/us_core_test_kit/reference_resolution_test.rb
```diff
@@ -12,15 +12,15 @@ def perform_reference_resolution_test(resources)
...
+      skip "Could not resolve and validate any Must Support references for #{unresolved_references_strings.join(', ')}"
     end
 
     def unresolved_references_strings
       unresolved_reference_hash =
         unresolved_references.each_with_object(Hash.new { |hash, key| hash[key] = [] }) do |missing, hash|
           hash[missing[:path]] << missing[:target_profile]
         end
-      unresolved_reference_hash.map { |path, profiles| "#{path}#{"(#{profiles.join('|')})" unless profiles.first.empty?}" }
+      unresolved_reference_hash.map { |path, profiles| "#{path} element: Reference(#{profiles.join('|')})" unless profiles.first.empty?}" }
```
Jammjammjamm, line 23: There's a syntax error here.

yunwwang: Forgot to push the last commit

PR #140, opened 2023-11-21
FI-2313 MS Slice Subelement Support

360dgries:
# Summary
Completion of [FI-2313](https://oncprojectracking.healthit.gov/support/browse/FI-2313)
- metadata extraction now collects both the tail end of slices and stores it under `slice_name` and stores the full id (not path) under `slice_id`
- areas that referenced slice[:name] correctly use slice_id 
- `fhir_resource_navigation` updated to be able to navigate slices using their discriminator.

### DIFF
--- a/lib/us_core_test_kit/fhir_resource_navigation.rb
+++ b/lib/us_core_test_kit/fhir_resource_navigation.rb
@@ -20,7 +20,6 @@ module USCoreTestKit
       return nil if element.nil?
 
       elements = Array.wrap(element)
-      path = (path.include?(':') && !path.include?('url')) ? path.gsub(/:.*\./, '.') : path
       if path.empty?
         unless include_dar
           elements = elements.reject do |el|
@@ -33,22 +32,20 @@ module USCoreTestKit
         return elements.first
       end
 
+      
 
       path_segments = path.split(/(?<!hl7)\./)
 
       segment = path_segments.shift.delete_suffix('[x]').gsub(/^class$/, 'local_class').to_sym
-
       no_elements_present =
         elements.none? do |element|
         child = get_next_value(element, segment)
-
         child.present? || child == false
       end
 
       return nil if no_elements_present
 
       remaining_path = path_segments.join('.')
-
       elements.each do |element|
         child = get_next_value(element, segment)
         element_found =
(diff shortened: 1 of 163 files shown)


lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
```diff
@@ -77,7 +77,8 @@ def must_support_pattern_slice_elements
       def pattern_slices
         must_support_pattern_slice_elements.map do |current_element|
           {
-            name: current_element.id,
+            slice_id: current_element.id,
+            slice_name: current_element.id.split(':')[1],
```
yunwwang, line 81: slice_name is coming from ElementDefinition.sliceName. Example
```
{
      "id" : "Coverage.class:group",
      "path" : "Coverage.class",
      "sliceName" : "group",
      ...
}
```

lib/us_core_test_kit/fhir_resource_navigation.rb
```diff
@@ -20,7 +20,7 @@ def find_a_value_at(element, path, include_dar: false)
       return nil if element.nil?
 
       elements = Array.wrap(element)
-
+      path = (path.include?(':') && !path.include?('url')) ? path.gsub(/:.*\./, '.') : path
```
yunwwang, line 23: This is not simple remove the slice name from the path. What need here is that if there is slice name in the middle, then navigate to the element at the specific path in a spcific slice.

Example, US Core Coverage.class
```
...
"class": [
  {
    "type": {
      "coding": [
        {
          "system": "http://example.com"
          "code": "local-code"
        }
      ]
    },
    "value": "abc"
  },
  {
    "type": {
      "coding": [
        {
          "system": "http://terminology.hl7.org/CodeSystem/coverage-class"
          "code": "group"
        }
      ]
    },
    "value": "def"
  },
    {
    "type": {
      "coding": [
        {
          "system": "http://terminology.hl7.org/CodeSystem/coverage-class"
          "code": "plan"
        }
      ]
    },
    "value": "ghi"
  },
]
```

then the path `class.group:value` navigates to the element with value "def"

360dgries: Ah, gotchya.  I had a misunderstanding of slices, and thought it could only have one.  Will patch this, thanks for the catch.

360dgries: I've added support for this -- currently it is looking through `type.coding.code` to pull the slice out.  I think the other slices I've looked at are also defined through this path, but let me know if you are aware of any that are not and this needs a more robust solution.

lib/us_core_test_kit/must_support_test.rb
```diff
@@ -15,15 +15,14 @@ def perform_must_support_test(resources)
       skip_if resources.blank?, "No #{resource_type} resources were found"
 
       missing_elements(resources)
-      missing_slices(resources)
-      missing_extensions(resources)
+      # missing_slices(resources)
+      # missing_extensions(resources)
 
-      handle_must_support_choices if metadata.must_supports[:choices].present?
+      # handle_must_support_choices if metadata.must_supports[:choices].present?
```
yunwwang, line 21: This should be added back to handle MustSupport choices

360dgries: Just went through and re-added any code I had commented out.  Should be good now, sorry about that!

PR #143, opened 2023-12-07
FI-2348 Add guard for fixed values on slice discriminators

360dgries:
Added a check in `handle_fixed_values` that prevents appending unnecessary information to slice subelement must supports.

No change in test output or rspec tests.  Elements that are not related to slices that have fixed values are still handled with no changes (see Observation.code for observation_occupation/metadata.yml)


360dgries:
I'm not sure if this needs to be more specific -- like if there are cases where a sub element of a slice has a fixed value, but it is not used within the discriminator of the slice.  I couldn't find a case in my inspection, but am definitely still learning the ins and outs of US Core structures.

PR #146, opened 2023-12-22
FI-2370: Add SMART App Launch

emichaud998:
# Summary
Added SMART App Launch tests to US Core test kit since support for SMART App Launch has been a US Core requirement since at least version 3.1.1, but SMART tests are not included in the test kit. Added Standalone and EHR launch tests as optional tests, each containing 4 test groups: SMART discovery (required), SMART ehr/standalone launch (required), SMART openid connect (optional), and smart token refresh (optional). Updated the US core presets so that it contains all the inputs for the SMART app launch tickets as well.

# Testing Guidance
Run the Inferno US Core test kit locally, and run the SMART App Launch tests using the Inferno Reference server preset, and make sure everything passes.

### DIFF
--- a/lib/us_core_test_kit/generator/templates/suite.rb.erb
+++ b/lib/us_core_test_kit/generator/templates/suite.rb.erb
@@ -76,22 +76,23 @@ module USCoreTestKit
 
 
       suite_option :smart_app_launch_version,
-      title: 'SMART App Launch Version',
-      list_options: [
-        {
-          label: 'SMART App Launch 1.0.0',
-          value: USCoreOptions::SMART_1
-        },
-        {
-          label: 'SMART App Launch 2.0.0',
-          value: USCoreOptions::SMART_2
-        }
-      ]
+        title: 'SMART App Launch Version',
+        list_options: [
+          {
+            label: 'SMART App Launch 1.0.0',
+            value: USCoreOptions::SMART_1
+          },
+          {
+            label: 'SMART App Launch 2.0.0',
+            value: USCoreOptions::SMART_2
+          }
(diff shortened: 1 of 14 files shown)


lib/us_core_test_kit/generator/templates/suite.rb.erb
```diff
@@ -71,11 +74,169 @@ module USCoreTestKit
...
+          label: 'SMART App Launch 1.0.0',
+          value: USCoreOptions::SMART_1
+        },
+        {
+          label: 'SMART App Launch 2.0.0',
+          value: USCoreOptions::SMART_2
+        }
+      ]
+
+      group do
```
Jammjammjamm, line 91: This is pretty sizeable, so let's move it into a separate file in `custom_groups`, give it an id of `us_core_smart_app_launch`, and then import it here.

Also, make sure any of the groups that aren't `group from: :some_id` have an id assigned which starts with `us_core` (`us_core_smart_standalone_launch_stu1`, etc.).

lib/us_core_test_kit/generator/templates/suite.rb.erb
```diff
@@ -71,11 +74,169 @@ module USCoreTestKit
...
+                client_id: { name: :ehr_client_id },
+                client_secret: { name: :ehr_client_secret },
+                received_scopes: { name: :ehr_received_scopes }
+              }
+            )
+          end
+        end
+      end
+
+      group do
```
Jammjammjamm, line 231: This will need an id which includes the us core version.

Jammjammjamm (approved):

PR #148, opened 2024-01-12
FI-2371: Add granular scopes tests

Jammjammjamm:
This branch adds testing for granular scopes.

During the normal FHIR API tests, requests for resources with required granular scopes are tagged with their resource type and search param names, e.g. `Condition?patient&category`. The granular scope tests then load up all of the requests from the normal FHIR API tests for a particular set of search params and repeat them, verifying that the resources returned are properly filtered according to the granted scopes.

To try it out:
- use US Core 6 or 7 & SMART App Launch 2
- run the normal US Core FHIR API tests
- run one of the granular scope launches & FHIR API w/Granular Scopes tests

Tasks in follow-up tickets:
- Add read interaction tests
- Add tests to verify that the required scopes were granted


yunwwang (approved):
Tested during Connectathon. Looks good

PR #149, opened 2024-01-28
FI-2231 Add searches for Practitioner test

yunwwang:
# Summary
This PR fixes JIRA ticket FI-2231. Practitioner is a delayed resource because it does not have patient search parameter and it is not an USCDI data class/data element. (g)(10) test kit excludes search tests for all delayed resources. 

To add search tests back for Practitioner in (g)(10) test kit, we split the original "delayed" resources into to part for (g)(10) test kit: delayed resources with searches and delayed resources without searches. The second category is identified by exclude_search_tests? metadata property.

# Change Log
* Expose is_delayed into metadata.yml (Note: I am still debating if this is necessary. I could remove this from metadata.yml if there're other opinions)
* Move some version specific constants to SpecialCase module
* Add Practitioner to SEARCHABLE_NON_USCDI_RESOURCES 
* Add logic of exclude_search_tests?

# Testing Guidance
This PR does not affect US Core Test Kit. Corresponding changes are also applied in (g)(10) test kit so that when (g)(10) loads US Core Test Suites, it uses the exclude_search_tests? property instead of delayed? property to decide if group's search tests shall be included/excluded.

### DIFF
--- a/lib/us_core_test_kit/generator/special_cases.rb
+++ b/lib/us_core_test_kit/generator/special_cases.rb
@@ -12,6 +12,41 @@ module USCoreTestKit
         'http://hl7.org/fhir/us/core/StructureDefinition/us-core-vital-signs'
       ].freeze
 
+      NON_USCDI_RESOURCES = {
+        'Encounter' => ['v311', 'v400'],
+        'Location' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'Organization' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'Practitioner' => ['v311', 'v400'],
+        'PractitionerRole' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'Provenance' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'RelatedPerson' => ['v501', 'v610', 'v700_ballot'],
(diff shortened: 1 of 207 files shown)


lib/us_core_test_kit/generator/special_cases.rb
```diff
@@ -12,6 +12,41 @@ module SpecialCases
...
+        'Location' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'Organization' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'Practitioner' => ['v311', 'v400'],
+        'PractitionerRole' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'Provenance' => ['v311', 'v400', 'v501', 'v610', 'v700_ballot'],
+        'RelatedPerson' => ['v501', 'v610', 'v700_ballot'],
+        'Specimen' => ['v610', 'v700_ballot']
+      }.freeze
+
+      SEARCHABLE_NON_USCDI_RESOURCES = {
```
Jammjammjamm, line 26: This naming is confusing because Practitioner is a USCDI resource is these versions. Why is Practitioner treated differently than Encounter? Aren't they both resources that weren't USCDI, but then became USCDI?

Jammjammjamm (approved):

PR #158, opened 2024-03-06
FI-2494 POST capability added to granular scopes

360dgries:
# Summary
Added inspection of request body if it is a POST request for query parameters.

Wondering about if we will ever expect more in the body than query parameters, i.e. should there be more thorough parsing/decision making here?

Also want to check that POST searches are made to an endpoint that ends in `\_search` every time?  The current spec test uses that endpoint, and I don't think it would impact the functionality, but the spec test wouldn't pass until I added that into the url (fhir_search must automatically add the route when it sees it's a POST?)

# Testing Guidance
Added a spec test that stubs a post request and returns correct resources, and expects test to pass.

### DIFF
--- a/lib/us_core_test_kit/granular_scope_checker.rb
+++ b/lib/us_core_test_kit/granular_scope_checker.rb
@@ -84,7 +84,6 @@ module USCoreTestKit
         granular_scopes.map do |scope|
           _, granular_scope = scope.split('?')
           name, value = granular_scope.split('=')
-
           {
             name:,
             value:
--- a/lib/us_core_test_kit/granular_scope_checker.rb
+++ b/lib/us_core_test_kit/granular_scope_checker.rb
@@ -16,7 +16,6 @@ module USCoreTestKit
       previous_request_resources.each do |previous_request, all_previous_resources|
         search_method = previous_request.verb.to_sym
         params = search_method == :get ? previous_request.query_parameters : Hash[URI.decode_www_form(previous_request.request_body)]
-        binding.pry if search_method == :post
         fhir_search(resource_type, params:, search_method:)
 
         found_resources =
--- a/lib/us_core_test_kit/granular_scope_checker.rb
+++ b/lib/us_core_test_kit/granular_scope_checker.rb
@@ -14,8 +14,8 @@ module USCoreTestKit
               "No #{resource_type} searches with search params #{search_param_names.join(' & ')} found"
 
       previous_request_resources.each do |previous_request, all_previous_resources|
-        params = previous_request.query_parameters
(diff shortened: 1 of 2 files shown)


lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -14,8 +14,8 @@ def run_scope_check_test
               "No #{resource_type} searches with search params #{search_param_names.join(' & ')} found"
 
       previous_request_resources.each do |previous_request, all_previous_resources|
-        params = previous_request.query_parameters
         search_method = previous_request.verb.to_sym
+        params = search_method == :get ? previous_request.query_parameters : JSON.parse(previous_request.request_body)
```
Jammjammjamm, line 18: The body of a POST search is `application/x-www-form-urlencoded` not JSON.

http://hl7.org/fhir/R4/search.html#Introduction

360dgries: Thanks, changed both the rspec test to encode it as such and the parsing here to treat it as `application/x-www-form-urlencoded` instead of json.

spec/us_core/granular_scope_checker_spec.rb
```diff
@@ -133,6 +133,7 @@ def self.metadata
           ).to_json
         )
       end
+
```
Jammjammjamm, line 136: Remove extra whitespace.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -14,8 +14,9 @@ def run_scope_check_test
               "No #{resource_type} searches with search params #{search_param_names.join(' & ')} found"
 
       previous_request_resources.each do |previous_request, all_previous_resources|
-        params = previous_request.query_parameters
         search_method = previous_request.verb.to_sym
+        params = search_method == :get ? previous_request.query_parameters : Hash[URI.decode_www_form(previous_request.request_body)]
+        binding.pry if search_method == :post
```
Jammjammjamm, line 19: Remove pry.

spec/us_core/granular_scope_checker_spec.rb
```diff
@@ -234,6 +235,48 @@ def self.metadata
...
+                { resource: matching_resource2.to_hash },
+                { resource: not_matching_resource.to_hash }
+              ]
+            ).to_json
+          )
+        end
+
+        it 'correctly pulls parameters from POST body' do
+          response_body =
+          FHIR::Bundle.new(
```
Jammjammjamm, line 264: Indent this.

PR #160, opened 2024-03-13
FI-2425: Ignore validator error for L/min unit on Pulse Ox profile

dehall:
# Summary
This PR ignores this validator error regarding the L/min unit on the Pulse Oximetry profile.
image

This error is a known issue since US Core 3.1.1 and is documented in the IG itself: https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-pulse-oximetry.html
> STU Note
> This profile is technically non-conformant with the base FHIR ver 4.0.1 vitals profile which defines a required binding that excludes the concept for L/min for `Observation.component.value(x)`. This is a known issue in and it means validation errors may occur when validating against this version of the US Core Pulse Oximetry Profile. A change request to correct this has been applied to the base FHIR ver 4.3.0 (“Release 4B”).

This error comes from the terminology lookup so it does not occur in g10.

# Testing Guidance
In production the reference server will fail this test on every version. Confirm the appropriate test passes for the Pulse Ox profile in each version.
Should the message filter be made more specific?


yunwwang (approved):
Tested with inferno service. This specific error is filtered as expected

PR #168, opened 2024-04-17
FI-2517: Migrate to HL7 validator wrapper

dehall:
# Summary
Migrates from the original Inferno validator wrapper to the new HL7 validator wrapper. There should be no visible difference in the results, and there should be no special customization or configuration needed to get anything working. However the validator UI is no longer available by default.

Unlike the g10 cutover which used an env var to toggle between the 2, this one is a complete transition, so for example things like the env vars are replaced instead of adding new ones.

Note that I did leave the old validator commented out in the docker-compose file and nginx config so that if someone wants to use the validator UI with US core preloaded, they can uncomment those. (Should I add a comment somewhere to explain that?)

There are a couple new error messages that don't show up on prod, but they do show up locally with the latest from `main` so I think the difference is just tx.fhir.org & caching related. The two messages I've seen are:
- Patient tests on v3.1.1 and 4.0.0: The Coding provided (http://terminology.hl7.org/CodeSystem/v3-NullFlavor#ASKU) was not found in the value set 'OMB Ethnicity Categories' 
- MedicationRequest tests on 4.0.0 and 7.0.0-ballot: Error from http://tx.fhir.org/r4: Unknown filter "tty = SCDG"

## Code changes
 - Bumped inferno-core to get the "don't run validator job in test" fix
 - Added hl7 validator to docker-compose and nginx config, commented out old validator items
 - Changed `validator` block in test suite template to `fhir_resource_validator`, added `igs`. Renamed env vars
 - Moved the suite `id` line to before the `fhir_resource_validator` block because the validator implicitly references the suite id so it needs to be set
 - Added new message to ignore `URL value '.*' does not resolve` - this happens when the validator sees a URL and is configured not to fetch it, and was ignored natively in the Inferno validator but the HL7 validator doesn't expose the ability to ignore that so it has to be ignored in the test kit
 - `validator.validate` calling the HL7 wrapper doesn't return OperationOutcome, but there is a function to convert it to one

# Testing Guidance
Since inferno-core was bumped, you may need to run db migrations.


yunwwang (approved):
Tested US Core v6.1.0 and v4.0.0. Results look good.

PR #170, opened 2024-04-19
Fi 2455 granular scope read tests

360dgries:
# Summary

Added read tests for Granular Scope groups.  

Tests attempt to find a resource from the first search of each profile that matches the given scope(s), each scope is tested individually.  After each scope has been checked, an attempt is made to find a resource that does not match any scopes.  A read is performed on this resource, and should not be returned.  Tests are skipped if any read cannot be populated from previous resources.

# Testing Guidance
Added Spec tests that cover most cases.  They behave as expected.

Running (US Core 7, Smart 2) against the reference server gives a 401, but I think that's a scope issue with the server -- by inspection:
The following request was made: `https://inferno.healthit.gov/reference-server/r4/Condition/84ab26cd-e0ea-4c8f-90fd-eddd939f3d9b`.

the read yielded:
```{
"resourceType": "OperationOutcome",
  "issue": [
    {
      "severity": "error",
      "code": "processing",
      "diagnostics": "Access to Resource Condition is restricted due to invalid scopes"
    }
  ]
}
```

but this resource has a category value of `encounter-diagnosis`, so it should fit the scopes and be returned properly?

The resource: 
```
{
      "fullUrl": "https://inferno.healthit.gov/reference-server/r4/Condition/84ab26cd-e0ea-4c8f-90fd-eddd939f3d9b",
      "resource": {
        "resourceType": "Condition",
        "id": "84ab26cd-e0ea-4c8f-90fd-eddd939f3d9b",
        "meta": {
          "versionId": "1",
          "lastUpdated": "2024-04-10T00:21:05.076+00:00",
          "profile": [
            "http://hl7.org/fhir/us/core/StructureDefinition/us-core-condition"
          ]
        },
        "clinicalStatus": {
          "coding": [
            {
              "system": "http://terminology.hl7.org/CodeSystem/condition-clinical",
              "code": "resolved"
            }
          ]
        },
        "verificationStatus": {
          "coding": [
            {
              "system": "http://terminology.hl7.org/CodeSystem/condition-ver-status",
              "code": "confirmed"
            }
          ]
        },
        "category": [
          {
            "coding": [
              {
                "system": "http://terminology.hl7.org/CodeSystem/condition-category",
                "code": "encounter-diagnosis",
                "display": "Encounter Diagnosis"
              }
            ]
          }
        ],
        "code": {
          "coding": [
            {
              "system": "http://snomed.info/sct",
              "code": "65363002",
              "display": "Otitis media"
            }
          ],
          "text": "Otitis media"
        },
        "subject": {
          "reference": "Patient/85"
        },
        "encounter": {
          "reference": "Encounter/c03951b3-00e5-4cb7-a271-805a15e6658b"
        },
        "onsetDateTime": "1947-11-01T00:11:45-05:00",
        "abatementDateTime": "1948-03-26T00:11:45-05:00",
        "recordedDate": "1947-11-01T00:11:45-05:00"
      }
    }
 ```

### CODE REVIEW

File: lib/us_core_test_kit/granular_scope_checker.rb [hunk 1]
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
       end
     end
 
+    def run_scope_check_read_test
```
Jammjammjamm, line 50: This needs to be a separate module.

lib/us_core_test_kit/generator/granular_scope_read_test_generator.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+
+      def array_of_strings(array)
+        quoted_strings = array.map { |element| "'#{element}'" }
+        "[#{quoted_strings.join(', ')}]"
+      end
+
+      def url_version
+        case group_metadata.version
+        when 'v3.1.1'
```
Jammjammjamm, line 85: Neither of these versions will be using granular scope tests.

lib/us_core_test_kit/generator/templates/granular_scope_read_test.rb.erb
```diff
@@ -0,0 +1,30 @@
+require_relative '../../../../generator/group_metadata'
+require_relative '../../../../granular_scope_checker'
+
+module USCoreTestKit
+  module <%= module_name %>
+    class <%= class_name %> < Inferno::Test
+      include USCoreTestKit::SearchTest
```
Jammjammjamm, line 7: We should try to find a way not to include this since it isn't a search test.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+      resource_scopes = granular_scopes.select {|scope| scope.include?(resource_type)}
+
+      load_previous_requests_for_reads
+      skip_if previous_requests.blank?,
+              "No #{resource_type} reads found"
+      previous_resources_for_reads = previous_request_resources.values.flatten
+
+      resource_specific_granular_scope_search_params.each do |scope| 
+        resource_matching_scope = previous_resources_for_reads.find {|prev_resource| search_param_value(scope[:name], prev_resource) == scope[:value]}
+        skip_if resource_matching_scope.nil?, "Unable to find any resources to match scope #{granular_scopes.find {|granular| granular.include?(scope[:name]) && granular.include?(scope[:value])}}"
```
Jammjammjamm, line 61: Pull that huge interpolation out into a variable.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+
+        fhir_read resource_type, resource_matching_scope.id
+
+        assert_response_status(200)
+        assert_resource_type(resource_type)
+        assert resource.id.present? && resource.id == resource_matching_scope.id, "Unable to read resource #{resource_matching_scope.id}, but it matches scopes"
```
Jammjammjamm, line 67: This error message doesn't seem to match the assertion.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+      end
+
+      nonmatching_resource = previous_resources_for_reads.find do |prev_resource|
+        resource_specific_granular_scope_search_params.all? { |scope| search_param_value(scope[:name], prev_resource) != scope[:value]}
+      end
+
+      skip_if nonmatching_resource.nil?, "Unable to find a resource that does not match scopes"
```
Jammjammjamm, line 74: This shouldn't be a skip. Not having resources that don't match shouldn't prevent a system from passing.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+      fhir_read resource_type, nonmatching_resource.id
+
+      assert_response_status(401)
```
Jammjammjamm, line 77: 401 is not the only possible response status.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+      assert_resource_type(:OperationOutcome)
```
Jammjammjamm, line 78: Servers are not required to return OOs on errors.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+      assert !resource.id.present?, "Resource #{nonmatching_resource.id} does not match the scopes, but was still read."
```
Jammjammjamm, line 79: I don't think this assertion makes sense.

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -47,6 +47,38 @@ def run_scope_check_test
...
+      assert granular_scopes.present?, "No granular scopes were received for #{resource_type} resources"
+      resource_scopes = granular_scopes.select {|scope| scope.include?(resource_type)}
+
+      load_previous_requests_for_reads
+      skip_if previous_requests.blank?,
+              "No #{resource_type} reads found"
+      previous_resources_for_reads = previous_request_resources.values.flatten
+
+      resource_specific_granular_scope_search_params.each do |scope| 
+        resource_matching_scope = previous_resources_for_reads.find {|prev_resource| search_param_value(scope[:name], prev_resource) == scope[:value]}
```
Jammjammjamm, line 60: `search_param_value` is for finding a single value to use for a search, not for finding all possible values which could match a search param.

lib/us_core_test_kit/granular_scope_checker.rb
360dgries: You mentioned that if the server does have resources that don't match, we should attempt a read and make sure it fails. 

1. If there is one that doesn't match, and it is read, does that prevent passing?
2. If there isn't one, what is the expected response by the test then?  Maybe an info statement that says it was unable to test the other direction?  I.e. how do we inform the user that it wasn't tested because no resources were found to supplement the test?

lib/us_core_test_kit/granular_scope_checker.rb
360dgries: This ties into the questions above -- if a resource is found that doesn't match any scopes, and it's [policy dependent on how a server responds](https://hl7.org/fhir/R4/security.html#AccessDenied:~:text=the%20actual%20result%20method%20used%20needs%20to%20be%20controlled%20by%20policy%20and%20context) to that, how do we assert it was not read?  My understading based on the link is the best way would be to confirm that status not 200, and if it is, check the bundle is empty, but not sure if the list at this link is exhaustive (it mentions these are only "typical" methods of handling).

lib/us_core_test_kit/granular_scope_checker.rb
```diff
@@ -100,6 +145,83 @@ def load_previous_requests
         load_tagged_requests(search_params_tag(search_params_as_hash))
           .sort_by { |request| request.index }
     end
+    
+    def load_previous_requests_for_reads
+      params = metadata.searches.first[:names]
+      search_params_as_hash = params.each_with_object({}) do |name, hash|
+        hash[name] = nil
+      end
+      @previous_requests ||=
```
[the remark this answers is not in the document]
360dgries: Are you saying that the previous granular scope searches should have already populated requests?  I included this in case there was ever only a read requirement, as well as to help with spec testing -- could also adjust `load_previous_requests` to take an input, and have searches pass in `search_param_names` and reads pass in the first search combination.  Let me know if I'm understanding the concern correctly.

lib/us_core_test_kit/granular_scope_checker.rb
360dgries: Took a try at generalizing this to being three modules, with one having the shared functionality.  Ended up being a bit of a timesink.  Sending more info on slack.

lib/us_core_test_kit/generator/granular_scope_read_test_generator.rb
360dgries: Ok, should I remove from granular search tests as well then?

lib/us_core_test_kit/generator/templates/granular_scope_read_test.rb.erb
360dgries: Removed.  Artifact from duplicating search template.

lib/us_core_test_kit/granular_scope_checker.rb
360dgries: Done.

lib/us_core_test_kit/granular_scope_checker.rb
360dgries: Replaced with `resource_match_search_values?`

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -0,0 +1,241 @@
...
+      end
+
+      paths.map { |path| path.delete_prefix("Resource.") }  
+    end
+
+    def unescape_search_value(value)
+      value&.gsub('\\,', ',')
+    end
+    
+    def resource_matches_param?(resource, search_param_name, escaped_search_value, values_found = [])
```
Jammjammjamm, line 127: I would not consider this or `element_has _valid_value?` to be stable. If they need to be used in multiple places, they should be extracted to a module and included in each place rather than duplicated.

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -0,0 +1,241 @@
...
+
+      nonmatching_resource = previous_resources_for_reads.find do |prev_resource|
+        resource_specific_granular_scope_search_params.all? do |scope| 
+          !resource_matches_param?(prev_resource, scope[:name], scope[:value])
+        end
+      end
+      if nonmatching_resource
+        fhir_read resource_type, nonmatching_resource.id
+        #TODO: Decide assertions for nonmatching resources
+        assert !resource.id.present?, "Resource #{nonmatching_resource.id} does not match the scopes, but was still read."
```
Jammjammjamm, line 40: I think we should verify that the read doesn't result in a 200, and doesn't return the type of resource requested.

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -0,0 +1,241 @@
...
+      @granular_scopes ||=
+        received_scopes
+          .split(' ')
+          .select do |scope|
+            (scope.start_with?("patient/#{resource_type}") || scope.start_with?("user/#{resource_type}")) &&
+              scope.include?('?')
+          end
+    end
+
+    def granular_scope_search_params
```
Jammjammjamm, line 78: If we need functionality across the read and search tests, it should go in a module that both use.

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -0,0 +1,241 @@
...
+
+        fhir_read resource_type, resource_matching_scope.id
+
+        assert_response_status(200)
+        assert_resource_type(resource_type)
+        assert resource.id.present? && resource.id == resource_matching_scope.id, "Unable to read resource #{resource_matching_scope.id}, but it matches scopes"
+      end
+
+      nonmatching_resource = previous_resources_for_reads.find do |prev_resource|
+        resource_specific_granular_scope_search_params.all? do |scope|
```
Jammjammjamm, line 33: Use `.none?` rather than `.all?` with a not.

Jammjammjamm (commented):
I get a purple error when I run this:

Screenshot 2024-05-21 at 10 20 30 AM

lib/us_core_test_kit/granular_scope_read_test.rb
360dgries: I think I grabbed this everywhere I could, created two new modules, one for specifically granular methods, and one of navigating resources as they are represented by Inferno('s models).  There is one redundant method, `search_params_tag`, but the method is a one-liner and is hopefully acceptable as is.

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -30,76 +38,19 @@ def run_scope_read_test
...
+        resource_specific_granular_scope_search_params.none? do |scope| 
+          resource_matches_param?(prev_resource, scope[:name], scope[:value])
         end
       end
       if nonmatching_resource
         fhir_read resource_type, nonmatching_resource.id
-        #TODO: Decide assertions for nonmatching resources
-        assert !resource.id.present?, "Resource #{nonmatching_resource.id} does not match the scopes, but was still read."
+        assert response[:status] != 200, "Server incorrectly responded with a successful status, read should fail due to scopes."
+        assert resource.resourceType != resource_type, "Server incorrectly returned a #{resource_type}, read should fail due to scopes"
```
Jammjammjamm, line 48: This isn't nil-safe

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -30,76 +38,19 @@ def run_scope_read_test
...
+        assert !resource.id.present?, "Server incorrectly returned resource #{nonmatching_resource.id}, read should fail due to scopes"
```
Jammjammjamm, line 49: I don't think this assertion makes sense and we should just get rid of it.

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -0,0 +1,73 @@
...
+        resource_matching_scope = previous_resources_for_reads.find do |prev_resource| 
+          resource_matches_param?(prev_resource, scope[:name], scope[:value])
+        end
+        skip_if resource_matching_scope.nil?, "Unable to find any resources to match scope #{current_scope}"
+
+        fhir_read resource_type, resource_matching_scope.id
+
+        assert_response_status(200)
+        assert_resource_type(resource_type)
+        assert resource.id.present? && resource.id == resource_matching_scope.id, "Unable to read resource #{resource_matching_scope.id}, but it matches scopes"
```
Jammjammjamm, line 37: This assertion message states that the resource matches the scope when the id doesn't match, but we haven't verified that, nor do we need to, because the id has to match.

There's no need to check that `resource.id` is present, because it wouldn't match the other resource's id if it weren't.

lib/us_core_test_kit/resource_checker.rb
```diff
@@ -0,0 +1,139 @@
+require_relative 'fhir_resource_navigation'
+
+module USCoreTestKit
+  module ResourceChecker
```
Jammjammjamm, line 4: Let's call it `ResourceSearchParamChecker`.

spec/us_core/granular_scope_read_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+            tags: ['Condition?patient&category'],
+            response_body: FHIR::Bundle.new(
+              entry: [
+                { resource: matching_resource.to_hash },
+                { resource: matching_resource2.to_hash }
+                ]
+            ).to_json
+          )
+        end
+        #TODO: Decide on how to test the reverse direction if unable to find a resource
```
Jammjammjamm, line 217: Remove commented out code.

360dgries: Removed for now.  Do we want spec tests that capture the cases of if there is no resource that mismatches the scopes?  We currently still let tests pass, but have an info statement claiming it was unable to test the case of trying to read a resource we know doesn't fit.

If we do want spec tests for this case, how do you check the info of a test in spec tests? I was trying to access messages, but it was showing up as empty even if the test hit the `info` method (as it does in this test that is commented out)

Jammjammjamm: I'm not sure why the result isn't including the messages, but you can get them like this:

```
Inferno::Repositories::Messages.new.messages_for_result("0cd47ecc-14a0-4904-bcf1-d7c640aab738")
```

This seems like a reasonably important message, so I think it makes sense to include a unit test for it.

PR #171, opened 2024-04-22
FI-2456: Verify granular scopes

Jammjammjamm:
# Summary
This branch adds tests to verify that the required granular scopes were granted.

# Testing Guidance
Run one of the granular scope launches against the reference server. When the reference server allows you to select which scopes to grant, deselect some of the granular scopes, and test will fail.


yunwwang:
Test result looks as expected.
image

yunwwang (approved):

PR #175, opened 2024-05-13
FI-2749 Screen Assessment

yunwwang:
# Summary

This PR fixes Github Issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/508.
This JIRA ticket https://jira.hl7.org/browse/FHIR-45319 only address US Core v6.1.0. The mapping for US Core v7 is more complicated and will be addressed by a separated HL7 FHIR ticket. At this time, US Core v7-ballot Test Suite will use the same resolution as US Core v6.1.0

# Changes:
* Add special test Screening Assessment Guidance to US Core v6.1.0 and v7.0.0-ballot Test Suite
* Add Unit tests

# Testing Guidance

Run US Core v6.1.0 Test Suite
Run Test 2.49 "Screening and Assessment Guidance"

### DIFF
--- a/lib/us_core_test_kit/custom_groups/screening_assessment_category_test.rb
+++ b/lib/us_core_test_kit/custom_groups/screening_assessment_category_test.rb
@@ -1,0 +1,90 @@
+require_relative '../search_test'
+
+module USCoreTestKit
+  class ScreeningAssessmentCategoryTest < Inferno::Test
+    include SearchTest
+
+    id :us_core_screening_assessment_category
+    title 'Server demonstrates support for the required Condition and Observation categories.'
+
+    input :patient_ids,
+          title: 'Patient IDs',
+          description: 'Comma separated list of patient IDs that in sum contain all MUST SUPPORT elements'
+
+    CONDITION_REQUIRED_CATEGORY = 'sdoh'
+
+    # Disable tagging requests since this test doesn't have properties defined.
+    # TODO: investigate whether we need to tag these for granular scope
+    # checking.
+    def tags(_params)
+      nil
+    end
+
+    def condition_category_found?(patient_id)
+      search_params = { patient: patient_id }
+      category_search_values = ['health-concern', 'problem-list-item']
+
+      category_search_values.any? do |category_search_value|
+        search_params[:category] = category_search_value
+        search_and_check_response(search_params, 'Condition')
+
+        resources = fetch_all_bundled_resources(resource_type: 'Condition')
+                    .select { |resource| resource.resourceType == 'Condition' }
+
+        resources.any? do |resource|
+          resource.category&.any? do |category|
+            category.coding&.any? { |coding| CONDITION_REQUIRED_CATEGORY == coding.code }
+          end
+        end
+      end
+    end
+
+    def observation_categories_found(patient_id)
+      search_params = { patient: patient_id, category: 'survey' }
+      search_and_check_response(search_params, 'Observation')
+
+      resources = fetch_all_bundled_resources(resource_type: 'Observation')
+                    .select { |resource| resource.resourceType == 'Observation' }
+
+      codes_found =
+        resources.flat_map do |resource|
+          codes = resource.category&.flat_map do |category|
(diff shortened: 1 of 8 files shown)


lib/us_core_test_kit/custom_groups/screening_assessment_category_test.rb
```diff
@@ -0,0 +1,90 @@
...
+    include SearchTest
+
+    id :us_core_screening_assessment_category
+    title 'Server demonstrates support for the required Condition and Observation categories.'
+
+    input :patient_ids,
+          title: 'Patient IDs',
+          description: 'Comma separated list of patient IDs that in sum contain all MUST SUPPORT elements'
+
+    CONDITION_REQUIRED_CATEGORY = 'sdoh'.freeze
```
Jammjammjamm, line 14: It's strange to me to use a different method for the required Condition and Observation categories.

lib/us_core_test_kit/custom_groups/screening_assessment_category_test.rb
```diff
@@ -0,0 +1,90 @@
...
+      pass if !missing_con_category && missing_obs_categories.empty?
+
+      message = 'Could not find these '
+
+      if missing_obs_categories.present?
+        message += "Observation categories: #{missing_obs_categories.join(', ')}"
+        message += ' and ' if missing_con_category
+      end
+
+      message += "Condition category: #{CONDITION_REQUIRED_CATEGORY}" if missing_con_category
```
Jammjammjamm, line 83: If no observation categories are missing, the error message would be `Could not find these Condition category:`, which isn't grammatically correct.

lib/us_core_test_kit/generated/v7.0.0-ballot/us_core_test_suite.rb
```diff
@@ -5,6 +5,7 @@
 require_relative '../../custom_groups/data_absent_reason_group'
 require_relative '../../custom_groups/smart_app_launch_group'
 require_relative '../../custom_groups/v7.0.0-ballot/smart_granular_scopes_group'
+require_relative '../../custom_groups/v6.1.0/screening_assessment_group'
```
Jammjammjamm, line 8: Isn't US Core 7 going to need a different version of this test because more things have been added to USCDI (treatment intervention preference and care experience preference)?

lib/us_core_test_kit/custom_groups/screening_assessment_category_test.rb
yunwwang: Condition has only one code while Observation has a set. I could try to combine them toegether.

Jammjammjamm: There's only one for now. It seems strange to me to maintain a separate code path for the special case of only one code, when there's a good chance there will be more than one code in the future.

lib/us_core_test_kit/generated/v7.0.0-ballot/us_core_test_suite.rb
Jammjammjamm: What about treatment intervention preference and care experience preference? Those are in USCDI but aren't really screening assesments.

yunwwang: There are separate profiles for these two. The value set [US Core Screening Assessment Observation Category](https://build.fhir.org/ig/HL7/US-Core/ValueSet-us-core-screening-assessment-observation-category.html) in US Core v7 is wrong. It contains two concepts which should not be included but lack of 'activity' and 'social-history' which should be included. I have raised [FHIR-46052](https://jira.hl7.org/browse/FHIR-46052) to fix that. The correct values, according to the narrative guidance, are:
image

yunwwang:
Another issue is that this change is not added into CCG yet. I am thinking to change this group to optional so it does not shown in g10. We can remove the optional tag after CCG is published. Thought?

Jammjammjamm (approved):

PR #183, opened 2024-06-18
FI-2818: Revert granular scope change

Jammjammjamm:
# Summary
This branch reverts #169, so granular scopes in US Core 6.1.0 will again be tested using two groups with the scopes split between the groups rather than testing support for all granular scopes within a single group.

# Testing Guidance
If you run US Core 6.1.0 with SMART App Launch 2, you should see the two granular scope groups using distinct sets of scopes.

### DIFF
--- a/lib/us_core_test_kit/custom_groups/smart_scopes_constants.rb
+++ b/lib/us_core_test_kit/custom_groups/smart_scopes_constants.rb
@@ -5,6 +5,7 @@ module USCoreTestKit
     SMART_GRANULAR_SCOPES_GROUP1 = {
       'v610' => [
           'patient/Condition.rs?category=http://terminology.hl7.org/CodeSystem/condition-category|encounter-diagnosis',
+          'patient/Condition.rs?category=http://hl7.org/fhir/us/core/CodeSystem/condition-category|health-concern',
           'patient/Observation.rs?category=http://terminology.hl7.org/CodeSystem/observation-category|laboratory',
           'patient/Observation.rs?category=http://terminology.hl7.org/CodeSystem/observation-category|social-history',
         ].map(&:freeze).freeze,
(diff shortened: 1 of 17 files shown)


lib/us_core_test_kit/custom_groups/smart_scopes_constants.rb
```diff
@@ -28,6 +23,13 @@ module SmartScopesConstants
     }.freeze
 
     SMART_GRANULAR_SCOPES_GROUP2 = {
+      'v610' => [
+          'patient/Condition.rs?category=http://hl7.org/fhir/us/core/CodeSystem/condition-category|health-concern',
```
yunwwang, line 27: How do you decide which group a category should be put into? For example, Condition's "health-concern" category is in group 1 for US Core 7-ballot but group 2 for US Core 6.

Jammjammjamm: We should probably keep them consistent, but really it's arbitrary.

yunwwang: good to know. I was wondering if the intention is to keep categories from the same profile together in one group or separate into two groups.

lib/us_core_test_kit/generator/suite_generator.rb
```diff
@@ -64,8 +64,6 @@ def ig_identifier
 
       def ig_link
         case ig_metadata.ig_version
-        when 'v6.1.0'
```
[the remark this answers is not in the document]
Jammjammjamm: I just didn't notice that this got reverted.

lib/us_core_test_kit/custom_groups/smart_scopes_constants.rb
Jammjammjamm: I don't think it really matters, but I'll go ahead and make it consistent with what is there for the 7 ballot.

yunwwang (approved):

PR #184, opened 2024-06-26
FI-2792 Create US Core 7 Test Suite

yunwwang:
# Summary

Add US Core 7 Test Suite
Add "Additional US Core Requirements" test to SMART Discovery test group
Remove US Core 7 Ballot Test Suite

# Testing Guidance

Run US Core 7 Test Suite.
Some tests may fail due to terminology issue from tx.fhir.server
Some tests may fail due to reference server data (tickets should be already created)


lib/us_core_test_kit/custom_groups/smart_standalone_launch_stu2_group.rb
```diff
@@ -1,12 +1,17 @@
...
     id :us_core_smart_standalone_launch_stu2
     title 'Standalone Launch'
 
     run_as_group
 
-    group from: :smart_discovery_stu2,
-          run_as_group: true
+    group from: :smart_discovery_stu2 do
+      run_as_group
+      test from: 'us_core_smart_well_known_capabilities'
```
Jammjammjamm, line 12: `scopes_supported` exists in SMART App Launch v1, so I'm not sure whether this test should exist there too.

Jammjammjamm (commented):
There's another spot where versions need updating:

https://github.com/inferno-framework/us-core-test-kit/blob/main/lib/us_core_test_kit/generator/search_test_generator.rb#L211-L218

yunwwang:
> There's another spot where versions need updating:
> 
> https://github.com/inferno-framework/us-core-test-kit/blob/main/lib/us_core_test_kit/generator/search_test_generator.rb#L211-L218

Moved both to Naming class

lib/us_core_test_kit/custom_groups/smart_standalone_launch_stu2_group.rb
yunwwang: This is the confusing part because all the scopes listed in that section are SMART v2 scopes. I am not sure if the intention is that only servers support SMART v2 are required to support `scopes_supported`. I have raised a US Core JIRA to clarify that. (https://jira.hl7.org/browse/FHIR-46279)

Jammjammjamm (commented):
The tests need to be regenerated.

lib/us_core_test_kit/custom_groups/smart_standalone_launch_stu2_group.rb
Jammjammjamm: These tests appear for all versions of US Core rather than only for US Core 7.

Jammjammjamm (approved):

PR #194, opened 2024-09-03
FI-3508 Retrieve primitive value from navigation result

yunwwang:
# Summary
This PR fixes github issue https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/557

# Testing Guidance
All unit tests pass


Jammjammjamm (approved):

Jammjammjamm:
We should add a similar case to the reference server to ensure there are no regressions for this.

PR #203, opened 2024-10-24
FI-3367 Re-generate so that read tests are after always after search tests

Shaumik-Ashraf:
# Summary

Putting this up as a draft PR if anyone wants to investigate test kit behavior with this setup. I'm not expecting this to be merged to main. I'll close this PR once the bug preventing Encounter group to pass on its own is fixed (FI-3367).

# Testing Guidance


Shaumik-Ashraf:
Ticket FI-3367 was originally created to unblock FI-2727, which is now unblocked via another solution. Closing this PR.

PR #204, opened 2024-10-28
Rename US Core v3.1.1 profile support test to resource support test

yunwwang:
# Summary

US Core v3.1.1 has a special CapabilityStatement resource support test which verifies that server suports at least Patient + 1 resources. For US Core v4.0.0, this requirement is changed to server supports US Core Patient + 1 US Core profiles. The orginal resource support test for US Core v311 test suite accidentally had the same id as the one in US Core v4+ test suite.

# Change Log:
* Rename v3.1.1's profile_support_test.rb to resource_support_test.rb
* Change the test id for resource_support_test to match version and purpose

# Testing Guidance
* All unit test shall pass
* Running US Core 3.1.1 test suite using default reference server. CapabilityStatement test group (test group 2.1) shall pass


Jammjammjamm (approved):
This is a breaking change since it changes a test id.

PR #209, opened 2024-11-06
FI-3133 Test name typo for granular scope tests.

yunwwang:
# Summary
This PR fixes FI-3133 Test name typo for granular scope tests.
The root is that both the granular scope class and granular scope template have trailing term "Tests".

This PR can be reviewed after SVAP 2024.

# Code change
* Remove the trailing "Tests" from title property
* Add misisng id into generator template.

# Testing Guidance
Verify 9.14.2.2 and 9.14.2.3 test does not have "Tests Tests" in their test title.

### DIFF
--- a/lib/us_core_test_kit/generated/v6.1.0/condition_granular_scope1_group.rb
+++ b/lib/us_core_test_kit/generated/v6.1.0/condition_granular_scope1_group.rb
@@ -13,7 +13,7 @@ require_relative './granular_scope_tests/condition/condition_granular_scope_read
 module USCoreTestKit
   module USCoreV610
     class ConditionGranularScope1Group < Inferno::TestGroup
-      title 'Condition Granular Scope Tests Tests'
+      title 'Condition Granular Scope Tests'
       short_description 'Verify support for the server capabilities required by the US Core Condition Encounter Diagnosis Profile.'
       description %(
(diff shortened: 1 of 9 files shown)


lib/us_core_test_kit/generator/templates/suite.rb.erb
```diff
@@ -99,7 +99,8 @@ module USCoreTestKit
       group from: :<%= smart_app_launch_id %><% if us_core_7_and_above? %>,
             required_suite_options: USCoreOptions::SMART_1_REQUIREMENT
       group from: :<%= smart_app_launch_id %>,
-            required_suite_options: USCoreOptions::SMART_2_REQUIREMENT
+            required_suite_options: USCoreOptions::SMART_2_REQUIREMENT,
+            id: :us_core_v700_smart_app_launch_stu2
```
Jammjammjamm, line 103: I don't think this id should have `v700` in it since it would be used in future versions too, right?

yunwwang: rebased to latest main which fixed this issue by replacing the fix id with 
```
id: :<%= smart_app_launch_id %>_stu2
```

Jammjammjamm (approved):

PR #222, opened 2025-01-09
FI-3338 Platform Deployment Spec Test with Metadata

Shaumik-Ashraf:
# Summary

See file diff for test kit metadata definition for testing inferno-framework/inferno-core#586


Shaumik-Ashraf:
Testing inferno-framework/inferno-core#586 was completed, and the required implementation details was carried over in #225

PR #223, opened 2025-01-10
FI-3642 Update docs publishing workflow to US Core test kit

arscan:
# Summary

This applies the improved docs publishing workflow that we have already incorporated into the g10 test kit.  It makes it easier to preview updates to the documentation that get published to the wiki.  I'd like to get this in here before i add one final PR for a last round of US Core Test Kit documentation updates.

# Testing Guidance
I do not believe testing is necessary as this was tested in the g10 repo.


arscan:
Merged because this is a repo configuration change and does not impact the US Core Test Kit software.

PR #228, opened 2025-02-12
FI-3586: Migrate to authinfo

Jammjammjamm:
# Summary
This branch updates the US Core Test Kit to use the updated SMART App Launch Test Kit with AuthInfo instead of OAuthCredentials.

# Testing Guidance
All tests and presets should work as normal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /data/*.db
 /data/*.db-journal
+/data/igs/*.tgz
 /data/redis/*.rdb
 /data/redis/*.aof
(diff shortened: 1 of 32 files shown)


yunwwang:
Found an interesting "bug". Here is how to reproduce it:

- select US Core v6.1 + SMART 2.0
- select test 1.3
- run with "confidential symmetric" authorization.
- After successful pass, click "run again", and select "confidential asymmetric" authorization
- the "Submit" button is disabled

I don't think this is a blocker for this PR.

PR #229, opened 2025-02-14
FI-3754 Default suites to use smart v2.0.0 in new metadata

arscan:
inferno.healthit.gov defaults to smart v2 for all suites in US Core.  See https://inferno.healthit.gov/test-kits/us-core/

This updates our suite options to be consistent with that so there is no behavior change when we migrate over to the new method of leveraging test kit metadata on platform apps.

This operates under the assumption that we want to use this version of smart as the default, but it seems reasonable to me.  Feel free to disagree though -- we don't have to merge this.

This includes functionality provided by the new v0.6.4 inferno core release, which is why the gemspec reflects that as a minimum version.

The only way to test right now is to launch the app and check the API responses that come back; SMART App Launch 2 should be the default for all US Core suites.  A later upgrade to inferno core will update the standalone app UI, but in the meantime we'll be updating platforms to read that so as to not loose existing functionality on inferno.healthit.gov in moving towards test kit metadata usage on the test kit pages.

Screenshot 2025-02-14 at 2 25 56 PM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     us_core_test_kit (0.10.0)
-      inferno_core (>= 0.6.1)
+      inferno_core (>= 0.6.4)
       smart_app_launch_test_kit (>= 0.5.0)
       tls_test_kit (~> 0.3.0)
 
@@ -137,7 +137,7 @@ GEM
     httpclient (2.8.3)
     i18n (1.14.7)
       concurrent-ruby (~> 1.0)
-    inferno_core (0.6.2)
+    inferno_core (0.6.4)
       activesupport (~> 6.1.7.5)
       base62-rb (= 0.3.1)
       blueprinter (= 0.25.2)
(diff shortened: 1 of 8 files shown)


Jammjammjamm (commented):
The documentation for the JSON API will need to be updated to reflect this.

Also, a boolean on the individual options is not the way we set defaults for radio inputs, where we set the default value on the field. It seems strange to use a different convention here.

arscan:
Oh, I didn't realize 2 things:
1. this was a feature we put into the jekyll platform templates but not back into core
2. you could just dump whatever you wanted into here and the dsl wouldn't reject anything it didn't know about

I just kinda backed into throwing this in there based on how we implemented the platform templates and didn't realize I was inadvertently stuffing non-standard content in there.

I'm up for whatever solution.  But we do need some way to set defaults on suite options based on what we wanted to accomplish from a ux perspective when building inferno.healthit.gov.  Does your approach require a core update?  If so, we need that asap as it would block transitioning inferno.healthit.gov to use the new templates.  (not a huge deal to not having this ability i guess, but it would be a loose end and i hate making backwards progress on UX before moving on to the next thing).

arscan:
Updated to incorporate new version of inferno core that can properly specify defaults.

Jammjammjamm (approved):

PR #233, opened 2025-02-26
FI-3765: Validation Message Filter Unit Test

emichaud998:
# Summary
Add unit test to ensure that the validation message filter for the BMI dataAbsentReason validation error is working as expected.

# Testing Guidance
Run spec tests and ensure they all pass

### DIFF
--- a/spec/us_core/validation_test_spec.rb
+++ b/spec/us_core/validation_test_spec.rb
@@ -13,27 +13,6 @@ RSpec.describe USCoreTestKit::ValidationTest do
 
   let(:patient_ref) { 'Patient/85' }
 
-  let(:operation_outcome_success) do
-    {
-      outcomes: [{
-        issues: []
-      }],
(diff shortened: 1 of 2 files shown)


yunwwang (approved):

spec/us_core/validation_test_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+        test_session_id: test_session.id,
+        name: name,
+        value: value,
+        type: runnable.config.input_type(name)
+      )
+    end
+    Inferno::TestRunner.new(test_session: test_session, test_run: test_run).run(runnable)
+  end
+
+  describe 'Observation BMI dataAbsentReason Validation Filter' do
```
yunwwang, line 51: You can move the default pass and error to a separate describe because these are shall always pass no matter what "new" filter we added.

PR #236, opened 2025-03-12
FI-3817: MedicationRequest Multiple_Or Search Fix

emichaud998:
# Summary

The MedicationRequest resource has two optional search requirements on `patient+intent+encounter` and `patient+intent+authoredon`. Previously, when testing it with reference server with patient 355 only, both test failed with errors:

```
Could not find plan values from intent in any of the resources returned for Patient/355
or
Could not find order values from intent in any of the resources returned for Patient/355
```

These searches are not the first search, so the tests use the result from first search to compile search values. From the first search, there is one MedicationRequest with `intent=order` and `encounter=Encounter/1`, and also a MedicationRequest with `intent=plan` and no encounter reference. The multiple_or search test therefore expected that patient 355 had MedicationRequests with `intent=order, plan` and `encounter=Encounter/1`, but it was not able to recognize that the only MedicationRequest with `intent=order` did not have an encounter reference. 

Added a fix for this problem by adding a filter to the list of existing values that the multiple_or search test compiles. This filter makes it so that a multiple_or parameter value will not be added to the list of existing values if the value comes from a resource that does not include all of the search parameters that are currently being searched by. So, in this example, it would not add `order` to the list of existing values for the `intent` search param, since the resource that contains`intent` with this value does not contain all of the search parameters for the `patient+intent+encounter` search (missing encounter).

# Testing Guidance
Run the MedicationRequest test with patient 355 and ensure the `patient+intent+encounter` and `patient+intent+authoredon`search tests pass. Also, run the spec tests and ensure those pass.

### DIFF
--- a/spec/us_core/search_test_spec.rb
+++ b/spec/us_core/search_test_spec.rb
@@ -70,7 +70,8 @@ RSpec.describe USCoreTestKit::SearchTest do
                 code: 'vital-signs'
               }
             ]
-          }],
+          }
+        ],
         code: [
           {
             coding: [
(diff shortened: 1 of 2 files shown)


spec/us_core/search_test_spec.rb
```diff
@@ -403,41 +409,62 @@ def scratch_resources
...
-      stub_request(:get, "#{url}/MedicationRequest?patient=#{patient_id}&intent=proposal,plan,order,original-order,reflex-order,filler-order,instance-order,option")
+      stub_request(:get, "#{url}/MedicationRequest?encounter=Encounter/#{encounter_id}&intent=proposal,plan,order,original-order,reflex-order,filler-order,instance-order,option&patient=#{patient_id}")
         .to_return(status: 200, body: bundle_2.to_json)
       result = run(multiple_or_search_test, patient_ids: patient_id, url: url)
 
       expect(result.result).to eq('fail')
       expect(result.result_message).to eq("Could not find order values from intent in any of the resources returned for Patient/#{patient_id}")
     end
+
+    it 'passes if multiple-or search test does not return all existing values if associated resources do not contain all search params' do
```
yunwwang, line 451: This unit test is not working. If I commented out this line in search_test
```
.filter { |resource| contains_all_search_params(resource, search_params) }
```
This unit test still pass which should fail.

PR #239, opened 2025-03-19
FI-3813: Use core Must Support assertion

dehall:
# Summary
This PR removes most of the Must Support Test internal logic in favor of calling the new core assertion `assert_must_support_elements_present`. The test uses the metadata that is already extracted and present.

NOTE: this does not completely rework the MS logic, this PR is aiming for a "minimal but clean" approach as a start to a more complete migration. A further rework would touch the MustSupportMetadataExtractor and leverage the core version of that class. That should be fully compatible since the core version was copied from this test kit, but there would need to be care in addressing the "uscdi_only" logic -- in core the metadata extractor was converted to take an optional "requirement extension URL" and so the field name in the resulting metadata is correspondingly generic.

Note also: the Da Vinci DTR and PDEX test kits include the MustSupportTest module, so by keeping the `perform_must_support_test` method with the same signature, those test kits should continue to work as-is with these updates. 
Eg:
https://github.com/inferno-framework/davinci-dtr-test-kit/blob/995cf57eeab016bd4438c28cbc909fc6c7c85987/lib/davinci_dtr_test_kit/profiles/encounter/encounter_must_support_test.rb#L3
https://github.com/inferno-framework/davinci-pdex-test-kit/blob/b1f94b955b726788368275fec31bcc09549743f4/lib/davinci_pdex_test_kit/pdex_payer_server/explanation_of_benefit/explanation_of_benefit_must_support_test.rb#L37

# Testing Guidance
Run the suites using the reference server preset. With default settings, everything should pass as usual. To get an example of a "failing" result, switch the "Patient IDs" setting from "85,355" to just "85" and re-run. In this case not everything will pass but results should match what's in production.

To test the DTR and PDEX test kits, add this branch of this test kit to the Gemfile:
```
gem 'us_core_test_kit',
    git: 'https://github.com/inferno-framework/us-core-test-kit.git',
    branch: 'fi-3813-ms-assertion'
```

Then `bundle install` and run the MS tests. (You may need to play with version numbers of some of the other dependencies unfortunately) All results locally should match what's on prod.


yunwwang (approved):

elsaperelli (approved):
Tested with DTR and looks good!

PR #240, opened 2025-03-21
FI-3823 Fix Granular Read Test fails when server returns 4xx response without FHIR OperationOutcome payload

yunwwang:
# Summary

This PR fixes Github Issue: https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/629

When server returns 4xx response without FHIR OperationOutcome resource in the payload, one of the assert which verifies the returned resources do not have the same resource_type fails because `fhir_model` could not serialize empty payload into a FHIR resource.

The solution is to catch NoMethodError error before doing the above assert testing.

# Testing Guidance
Two new unit tests are added for the use cases when no body returned, and empty string "" body returned.
All unit tests shall pass.


lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -37,19 +37,26 @@ def run_scope_read_test
...
         assert (response && response[:status]) != 200, "Server incorrectly responded with a successful status, read should fail due to scopes."
-        assert (resource && resource.resourceType != resource_type), "Server incorrectly returned a #{resource_type}, read should fail due to scopes"
+
+        begin
+          res = resource
+        rescue NoMethodError
+          res = nil
+        end
+
+        assert (!res || (resource && resource.resourceType != resource_type)), "Server incorrectly returned a #{resource_type}, read should fail due to scopes"
```
Jammjammjamm, line 54: I find this assertion incomprehensible.

yunwwang: rewrite the logic to be eqivalent to the original one.

lib/us_core_test_kit/granular_scope_read_test.rb
```diff
@@ -37,19 +37,26 @@ def run_scope_read_test
...
         assert (response && response[:status]) != 200, "Server incorrectly responded with a successful status, read should fail due to scopes."
-        assert (resource && resource.resourceType != resource_type), "Server incorrectly returned a #{resource_type}, read should fail due to scopes"
+
+        begin
+          res = resource
+        rescue NoMethodError
+          res = nil
+        end
+
+        assert (!res || res.resourceType == 'OperationOutcome'), "Read shall fail due to scopes. Server shall return 4xx status code with an optional OperationOutcome payload."
```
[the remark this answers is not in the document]
yunwwang: It is already there: https://github.com/inferno-framework/us-core-test-kit/blob/9ced212c3f8a31a71cc583048d4f9ac8ef4a9e8e/spec/us_core/granular_scope_read_spec.rb#L155

Jammjammjamm (approved):

PR #243, opened 2025-04-21
US Core Client read/search tests

holmesie:
# Summary

Needs: https://github.com/inferno-framework/inferno-reference-server-data/pull/2

Adds us core client read/search test suite generator.

```
bundle exec rake us_core:generate
```

# Testing Guidance

See lib/us_core_test_kit/client/docs/us_core_client.rb for a test script that can be run against the suite


Rakefile
```diff
@@ -24,4 +24,13 @@ namespace :us_core do
 
     USCoreTestKit::Generator.generate
   end
+
+  namespace :client do
```
Jammjammjamm, line 28: `rake us_core:generate` should generate everything. We shouldn't have to run two tasks.

lib/us_core_test_kit/client/test_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      def filter_requests_by_resource_id(requests, resource_id)
+        requests.select do |request|
+          request.url.split('/').last.split('?').first&.casecmp?(resource_id)
+        end
+      end
+
+      def filter_requests_by_search_parameters(requests, search_parameters)
+        requests.select do |request|
+          puts request.url
```
Jammjammjamm, line 20: No `puts` allowed.

.env.development
```diff
@@ -1,3 +1,5 @@
 FHIR_RESOURCE_VALIDATOR_URL=http://localhost/hl7validatorapi
 INFERNO_HOST=http://localhost:4567
-REDIS_URL=redis://localhost:6379/0
\ No newline at end of file
+REDIS_URL=redis://localhost:6379/0
+FHIR_REFERENCE_SERVER=http://localhost:8080/reference-server/r4
```
Jammjammjamm, line 4: These environment variables and the requirements to run the client suites should be documented in the README.

Jammjammjamm (commented):
I don't have time for a full review, just a couple quick comments.

yunwwang:
spotty checked some tests. This is a good start. Is there plan to implement these search requirements either in this implementer or in the future version?
* date comparator
* multiple-OR search

lib/us_core_test_kit.rb
```diff
@@ -4,4 +4,5 @@
 require_relative 'us_core_test_kit/generated/v5.0.1/us_core_test_suite'
 require_relative 'us_core_test_kit/generated/v6.1.0/us_core_test_suite'
 require_relative 'us_core_test_kit/generated/v7.0.0/us_core_test_suite'
-
+require_relative 'us_core_test_kit/client/generated/v6.1.0/us_core_client_test_suite'
+require_relative 'us_core_test_kit/client/generated/v7.0.0/us_core_client_test_suite'
```
emichaud998, line 8: Is there a reason why we are generating client tests for versions of us core that we are not including here?

lib/us_core_test_kit/client/generated/v7.0.0/wait_group.rb
```diff
@@ -0,0 +1,525 @@
...
+    * us-core-client-tests-patient
+  * searches:
+    * _id
+    * identifier
+    * name
+    * birthdate + name
+    * family + gender
+    * birthdate + family
+    * gender + name
+    * death-date + family
```
emichaud998, line 59: It may be worth listing the search param values next to each of the search parameters listed for each resource to make it easier for the user to make requests

emichaud998 (commented):
Ran through the tests and it looks really good! Just a few minor comments/suggestions

lib/us_core_test_kit/client/generated/v7.0.0/wait_group.rb
holmesie: The _search_ tests are basically just checking that the request with the correct params actually happened, not that something was returned. My hope was that the _read_ tests were sufficient to do any "received and successfully processed" attestations. Otherwise this might get a little messy with things being auto-generated, e.g. would probably require a lookup table of profile->param values or something to pull from in order to populate the right things. Alternatively maybe the generator takes in the actual JSON bundle that gets added to the reference server and pulls the right things directly out of there - so the only manual work required is maintaining that bundle. That might need to be follow up work though.

lib/us_core_test_kit.rb
holmesie: No, I've added the additional ones.

lib/us_core_test_kit/client/generated/v3.1.1/us_core_client_test_suite.rb
```diff
@@ -0,0 +1,240 @@
...
+- Testing a client's ability to perform searches using the SHALL conformance search parameters
+summarized in the US Core Capability Statement.
+
+The current version of this test suite does not support:
+- Testing searches with/via:
+  - date comparator
+  - multiple-OR
+  - _revInclude
+  - The client SHALL provide values precise to the day for elements of datatype date and to the second + time offset for elements of datatype dateTime.
+- Other must support requirements not outlined above.
```
yunwwang, line 170: - The Must Support Conformance Requirements for clients/requestors specified in [US Core IG v7.0.0](https://build.fhir.org/ig/HL7/US-Core/must-support.html#must-support) (note: update the US Core version and link for other test suites)

lib/us_core_test_kit/client/generated/v7.0.0/wait_group.rb
emichaud998: Good point, the fact that these are auto-generated would make this a little messy. Maybe this is just something to think about for the future but I think it is worth holding off for now

emichaud998:
Ran through all the tests and everything is working the way I would expect, Yunwei would know more from the requirements side if anything is missing but this looks good to me!

emichaud998 (approved):

lib/us_core_test_kit/client/generated/v3.1.1/us_core_client_test_suite.rb
yunwwang: Change "- Other must support requirements not outlined above." to
"The Must Support Conformance Requirements for clients/requestors specified in US Core IG v3.1.1"

(note: make similar update for other the US Core test suites)

yunwwang (approved):

PR #245, opened 2025-05-14
Add Client Auth framework support

karlnaden:
# Summary

Adds SMART and UDAP simulations to the client test suites.

# Testing Guidance

- Follow the instructions in the suite documentation to run the client suites against the server suites using a SMART public client. 
- Test Confidential Symmetric and Asymmetric clients using the same approach, but with the additional inputs needed for the authentication options
  - Symmetric: In the server tests, change the "Auth Type" input to "Confidential Symmetric" and add `SAMPLE_SECRET` as to the Client Secret input in both sessions
  - Asymmetric: In the server tests, change the "Auth Type" input to "Confidential Asymmetric" and in the client tests, add 'http://localhost:4567/custom/smart_stu2_2/.well-known/jwks.json' for the JWKS input.
- Test UDAP using the same approach, but getting an access token using the UDAP security tests instead of the standalone launch group
  - put `http://localhost:4567/custom/udap_security/fhir` as the UDAP Client URI in the client tests and start them
  - Start a UDAP Security Server test session and select the "Demo: Run against the UDAP Security Client Suite" preset. Update the **FHIR Server Base URL** input to `http://localhost:4567/custom/us_core_client_v700/fhir` (or the corresponding one for the target US Core version) and run the tests. Discovery and registration will have expected failures, but the token exchange will succeed (you'll need to click to authorize). Also click through the client session wait dialogs until it is waiting for US Core requests.
  - Pull the access token out of the outputs of the "Token exchange response body contains required information encoded in JSON" test. Then, start a US Core server session, select the "Demo:" preset and immediately execute the "US Core FHIR API" group, putting the UDAP access token in the inputs and run the tests. Once those are done, continue the client tests. The UDAP auth interaction group should succeed.

NOTE: test on a variety of US Core versions - always with client and server using the same version.  But you shouldn't have to test all combinations since the code is generated and something that works in one US Core version should work in the others.


holmesie (approved):

lib/us_core_test_kit/client/generator/suite_generator.rb
```diff
@@ -77,21 +77,98 @@ def description
...
+            - SMART [App Launch flow](https://hl7.org/fhir/smart-app-launch/STU2.2/app-launch.html), or
+            - UDAP [Consumer-Facing flow](https://hl7.org/fhir/us/udap-security/STU1/consumer.html).
+
+            When creating a test session, select the Client Security Type corresponding to an
+            authentication approach supported by the client. Then, start by running the "Client Registration"
+            group which will guide you through the registration process, including what inputs to provide.
+            See the *Auth Configuration Details* section below for details.
+
+            Once registration is complete, run the "Read & Search" group to have Inferno wait for US Core
+            read and search requests from the client, return the requested CARIN
```
emichaud998, line 91:
```suggestion
            read and search requests from the client, return the requested US Core
```

emichaud998 (approved):
One super tiny comment, otherwise this looks good and everything is working as expected!

PR #247, opened 2025-05-15
FI-3948 Prefers search with codings from Well-Known code system

yunwwang:
# Summary

This PR fixes JIRA Ticket FHIR-3948 Prioritize searching based on standard codes.

CodeableConcept may have multiple codings representing the 

# Change Log:
* Add WellKnownCodeSystems module
* Reorganize search test to use WellKnownCodeSystems to find the first coding from a well-known code system. If not, then use the first coding met.

# Testing Guidance


lib/us_core_test_kit/fhir_resource_navigation.rb
```diff
@@ -19,7 +19,7 @@ def resolve_path(elements, path)
       end.compact
     end
 
-    def find_a_value_at(element, path, include_dar: false)
+    def find_a_value_at(element, path, include_dar: false, prefer_well_known_code_sytem: false)
```
emichaud998, line 22:
```suggestion
    def find_a_value_at(element, path, include_dar: false, prefer_well_known_code_system: false)
```

lib/us_core_test_kit/search_test.rb
```diff
@@ -574,6 +576,15 @@ def fetch_and_assert_all_bundled_resources(
         fetch_all_bundled_resources(resource_type:, bundle:, reply_handler: reply_and_assert_handler, max_pages:, additional_resource_types:, tags:)
     end
 
+    def prefer_well_known_code_sytem(element, include_system)
```
emichaud998, line 579:
```suggestion
    def prefer_well_known_code_system(element, include_system)
```

emichaud998 (commented):
Typo comments, otherwise this looks good. Should we consider adding a spec test for these changes?

lib/us_core_test_kit/well_known_code_systems.rb
```diff
@@ -0,0 +1,21 @@
+module USCoreTestKit
+  module WellKnownCodeSystems
+    HL7_CODE_SYSTEMS = [
+      'http://terminology.hl7.org/CodeSystem',
+      'http://hl7.org/fhir/us/core/CodeSystem'
+    ].freeze
+
+    WELL_KNOWN_CODE_SYSTEMS = [
+      'http://loinc.org',
+      'http://snomed.info/sct'
```
Jammjammjamm, line 10: What's the reasoning behind only including loinc and snomed here rather than all of the code systems referenced in US Core?

yunwwang: These are used in those elements targeted by mandatory searche parameters. Other code systems are either not targeted by seach parameters or targetd by optional search parameters which are not tested in g10 certification.

lib/us_core_test_kit/fhir_resource_navigation.rb
```diff
@@ -19,7 +19,7 @@ def resolve_path(elements, path)
       end.compact
     end
 
-    def find_a_value_at(element, path, include_dar: false)
+    def find_a_value_at(element, path, include_dar: false, prefer_well_known_code_system: false)
```
Jammjammjamm, line 22: It doesn't look like this is actually used.

lib/us_core_test_kit/well_known_code_systems.rb
yunwwang: Another option is to use the code systems defined at https://terminology.hl7.org/external_terminologies.html.

Jammjammjamm (approved):

emichaud998 (approved):

PR #250, opened 2025-05-22
Improve Proxied Responses

karlnaden:
# Summary

Previously, the client tests returned only the status and response body when proxying requests to the inferno reference server. Now, it returns modified headers and also updates links to use the test suite's host and base fhir path.

# Testing Guidance

Run the client tests, make a FHIR request, and verify that responses returned by the client suite include appropriate content-type headers (e.g., application/fhir+json). Minimal test:
- make sure that the inferno-reference-server is running on port 8080
- start an instance of the US Core client v7.0.0 suite.
- select the "Demo: run against the server tests" preset
- Run all tests, clicking through the first wait dialog to confirm setup.
- in another tab, start an instance of test US Core server v7.0.0 suite
- select the "Demo: run against the client tests" preset
- Run group "3.5 Standalone Launch" to get an access token
- Run group "4.2 Patient Tests" to make a request
- look at the search requests in test 4.2.01 and the read request in test 4.2.09 and confirm the response and the response headers look right

### DIFF
--- a/lib/us_core_test_kit/client/generator/templates/read_endpoint.rb.erb
+++ b/lib/us_core_test_kit/client/generator/templates/read_endpoint.rb.erb
@@ -34,10 +34,6 @@ module USCoreTestKit
           end
         end
 
-        def resource_id
-          request.params[:resource_id]
-        end
-
         def resource_type
           request.params[:resource_type]
         end
--- a/lib/us_core_test_kit/client/generator/templates/read_endpoint.rb.erb
+++ b/lib/us_core_test_kit/client/generator/templates/read_endpoint.rb.erb
@@ -2,12 +2,14 @@
 
 require_relative '../../server_proxy'
 require_relative 'tags'
+require_relative 'urls'
 
 module USCoreTestKit
   module Client
     module <%= module_name %>
       class ReadEndpoint < Inferno::DSL::SuiteEndpoint
         include ServerProxy
+        include URLs
 
         def test_run_identifier
           UDAPSecurityTestKit::MockUDAPServer.issued_token_to_client_id(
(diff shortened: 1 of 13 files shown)


lib/us_core_test_kit/client/generator/templates/read_endpoint.rb.erb
```diff
@@ -16,13 +18,7 @@ module USCoreTestKit
...
         def make_response
-          server_response = proxy_request
-          response.status = server_response.status
-          response.body = server_response.body
-        end
-
-        def proxy_request
-          proxy_client.get("#{resource_type}/#{resource_id}", request_params)
+          build_proxied_read_response
         end
```
emichaud998, line 22: Since this `proxy_request` is being removed, I believe we no longer need the `resource_id` or `request_params` functions included in this file

lib/us_core_test_kit/client/server_proxy.rb
```diff
@@ -20,6 +20,56 @@ def request_headers
...
+                        else
+                          server_response.body
+                        end
+        response.headers.merge!(server_response.headers)
+        remove_transfer_encoding_and_content_length_header(response.headers)
+      end
+
+      def remove_transfer_encoding_and_content_length_header(headers)
+        headers.delete('transfer-encoding') if headers['transfer-encoding'].present?
+        headers.delete('Content-Length') if headers['Content-Length'].present?
```
emichaud998, line 54: It looks like content length might not be getting removed correctly? I can see it included in the response headers:

Screenshot 2025-05-22 at 11 22 12 AM

emichaud998 (commented):
Two small comments, otherwise everything else looks good

lib/us_core_test_kit/client/server_proxy.rb
karlnaden: it gets re-generated.  When not removed first, Inferno gets an unhappy purple error

PR #259, opened 2025-06-27
FI-4176: Update implementation docs

Jammjammjamm:
This branch updates the documentation on unusual implementation details to remove examples which have been removed. This branch also removes `fhir_resource_navigation` and `primitive_type` which have moved into core.


arscan (approved):
🎉

PR #268, opened 2025-07-15
FI-4221 Remove codes other than sdoh from MustSupport for Condition.category:screening-assessment

yunwwang:
# Summary

US Core v8 Condition Problems and Health Concerns Profile has a guidance that

The category of "problem-list-item" or "health-concern" is required, and, at a minimum, Certifying Systems SHALL support, a category of "sdoh"

This PR removes codings other than `sdoh` from the MustSupport metadata

# Testing Guidance

### DIFF
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor_us_core_8.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor_us_core_8.rb
@@ -27,6 +27,7 @@ module USCoreTestKit
         add_must_support_choices
         us_core_6_extractor.remove_practitioner_address
         remove_patient_and_encounter_interpreter_extension
+        apply_condition_sdoh_guaidance
       end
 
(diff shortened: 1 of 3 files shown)


lib/us_core_test_kit/generator/must_support_metadata_extractor_us_core_8.rb
```diff
@@ -51,6 +52,19 @@ def add_must_support_choices
           must_supports[:choices].concat(more_choices)
         end
       end
+
+      # US Core v8 Condition Problems and Health Concerns Implementation Guidance:
+      # The category of "problem-list-item" or "health-concern" is required, and, at a minimum,
+      # Certifying Systems SHALL support, a category of "sdoh"
+      def apply_condition_sdoh_guaidance
```
Jammjammjamm, line 59: Typo, should be `guidance`.

PR #275, opened 2025-07-30
FI-4268 Allow client read test to accept multiple resource ids

yunwwang:
# Summary

Observation Clinical Test Result profile is a parent profile for Observcation Lab Result profile. When client sends query for Clinical Test Result with patient ID, Lab Result Observation are also returned. So when client run read test, the resource id could be either `observation-clinical-test-result`, or `observation-lab`.

This PR allows client read test to accept multiple resource ids

This PR also fixed some incorrect resource ids mentioned in the group discription text which does not affect actual testing.

# Testing Guidance

Run US Core Client test for v6.1.0, v7.0.0, or v8.0.0. 
Verify Observation Clinical Test Result group passed


karlnaden (approved):
one minor documentation clean-up suggestion, otherwise, good

PR #282, opened 2026-01-07
ID-30: Handle relative bundle references during granular scope tests

ljtucker:
# Fix pagination URL comparison for granular scope validation

## Problem
The granular scope validation logic was failing to correctly group multi-page search results when FHIR bundles contained relative pagination URLs. The `previous_request_resources` method compared absolute request URLs directly with pagination URLs from bundle link elements using string equality. Since FHIR servers can return either relative URLs (e.g., `/Observation?...&_offset=10`) or absolute URLs in pagination links, the direct string comparison would fail for relative URLs, causing pages 2, 3, etc. to be treated as separate searches rather than continuations of the first request.

This resulted in test failures with errors like:
> Resources with the following ids were received when using granular scopes, but not when using resource-level scopes

## Solution
Added URL normalization to ensure consistent comparison between request URLs and pagination URLs:

**Added `resolve_url` method (granular_scope.rb lines 42-59)**
- Parses URLs to detect if they're already absolute (have a scheme)
- For relative URLs, resolves them against the **configured FHIR base URL** (`url` input) per FHIR spec
- **Special URI.join handling**: Adds trailing slash to base URL and strips leading slash from relative URLs to prevent URI.join from replacing the entire path (e.g., prevents `/fhir` from being stripped when resolving `/Observation`)
- Returns URLs in absolute form for accurate comparison
- Handles `URI::InvalidURIError` gracefully

**Updated `previous_request_resources` method (granular_scope.rb line 70)**
- Resolves pagination URLs using `resolve_url` before comparing with request URLs
- Properly groups all pages of a paginated search result together

## Testing
Added comprehensive test coverage in [granular_scope_search_spec.rb](spec/us_core/granular_scope_search_spec.rb):

✅ **Pagination with relative URLs** (lines 434-477)
- Tests proper grouping of multi-page results
- Tests failure detection when pages are missing

✅ **Pagination with absolute URLs** (lines 479-570)
- Tests proper grouping with fully qualified URLs
- Tests failure detection with incomplete pagination

✅ **Multiple separate searches** (lines 572-650)
- Ensures different searches aren't incorrectly grouped as pagination

**Test Infrastructure Fix:**
- Discovered that nested `repo_create(:result, ...)` calls within request definitions were creating separate test scenarios, causing only the last request to be found
- Fixed by using RSpec's memoized `let(:repo_create_output)` to create a single shared result object that both page requests reference
- This ensures both requests are associated with the same result ID and loaded together by `load_previous_requests`

All 27 tests pass (19 search + 8 read), confirming the fix handles both relative and absolute pagination URLs correctly while maintaining backward compatibility.

## Impact
- ✅ Fixes granular scope validation for servers that use relative pagination URLs
- ✅ Maintains compatibility with servers using absolute pagination URLs  
- ✅ Properly handles base URL paths (e.g., `/fhir/`) without stripping them during URL resolution
- ✅ No breaking changes to existing tests or functionality
- ✅ More robust URL handling with proper parsing and error handling

### DIFF
--- a/spec/us_core/granular_scope_search_spec.rb
+++ b/spec/us_core/granular_scope_search_spec.rb
@@ -256,7 +256,21 @@ RSpec.describe USCoreTestKit::GranularScopeSearchTest, :runnable do
           expect(result.result).to eq('pass')
         end
 
-         context 'with POST requests' do
+        it 'ignores resource-level scopes provided' do
+          stub_request(request.verb.to_sym, request.url.split('?').first)
+            .with(query: request.query_parameters)
+            .to_return(body: FHIR::Bundle.new(
+                                entry: [
+                                  { resource: matching_resource.to_hash },
+                                  { resource: matching_resource2.to_hash }
+                                ]
+                              ).to_json)
+          
+          result = run(granular_scope_test, url:, patient_ids:, received_scopes: "patient/Observation.rs #{received_scopes}")
+          expect(result.result).to eq('pass')
+        end
+
+        context 'with POST requests' do
           let!(:post_request) do
             repo_create(
               :request,
@@ -426,7 +440,6 @@ RSpec.describe USCoreTestKit::GranularScopeSearchTest, :runnable do
           expect(result.result).to eq('fail')
           expect(result.result_message).to eq(expected_message)
         end
-
       end
 
       context 'with paginated results using relative URLs' do
--- a/lib/us_core_test_kit/custom_groups/v4.0.0/document_reference_custodian_test.rb
+++ b/lib/us_core_test_kit/custom_groups/v4.0.0/document_reference_custodian_test.rb
@@ -29,7 +29,7 @@ module USCoreTestKit
 
           if provenances.present?
             has_agent = provenances.any? do |provenance|
-              provenance.target.any? { |target| target.reference.end_with?("DocumentReference/#{docref.id}") } &&
+              provenance.target.any? { |target| target.reference.present? && target.reference.end_with?("DocumentReference/#{docref.id}") } &&
               provenance.agent.any? { |agent| agent.who.present? || agent.onBehalfOf.present? }
             end
           end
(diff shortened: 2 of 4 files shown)


spec/us_core/granular_scope_search_spec.rb
```diff
@@ -429,6 +429,261 @@ def self.metadata
...
+
+        it 'properly groups paginated resources from relative URLs' do
+          # Stub page 1 with relative next link
+          stub_request(:get, 'http://example.com/fhir/Observation')
+            .with(query: hash_including('patient' => 'PATIENT_ID', 'category' => 'http://terminology.hl7.org/CodeSystem/observation-category|survey'))
+            .to_return(
+              body: FHIR::Bundle.new(
+                type: 'searchset',
+                link: [
+                  { relation: 'next', url: '/Observation?patient=PATIENT_ID&category=http://terminology.hl7.org/CodeSystem/observation-category%7Csurvey&_offset=10' }
```
karlnaden: I think this should fail because "http://example.com/fhir/Observation" will be the base_url and you'll only bring over "http://example.com" from it (scheme and host), so the normalized next link would be "http://example.com/Observation?patient=..." which is missing the first "fhir" path.

I'm not quite sure why but your test cases are testing the situation that you think they are. If I revert the logic and leave the new unit tests in place, they still pass. I tried putting breakpoints within the new code and it never reached past the first line of `resolve_url`. I worked on setting up a method to reproduce the situation in a running server (using Mockoon - can give more details tomorrow) and was able to get the same error with the updated code. So we need to figure out what's going wrong with the test case and fix it.

ljtucker: I have updated the spec tests to share the same results instance, which should ensure the pagination tests respect both pages

karlnaden (approved):

arunkumarhk (approved):

Issue #283, opened 2026-01-09
MustSupport metadata extractor has a bug in method type_slice


yunwwang:
in the [type_slice method](https://github.com/inferno-framework/us-core-test-kit/blob/00513fad7d7cab0ce27a185fc74f4d8a7e33a69e/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb#L88)

variable `elements` is not defined. The variable shall be used here is `profile_elements`. Since no US profile has type slice with path, this line is never hit.

Suggest to replace `elements` with the correct variable `profile_elements`, and add `path: type_path` to the returned discriminator to make the function complete.

This fix does not affect any US Core testing since US Core does not have type slice with path.

I have logged a feature improvement for [inferno_core](https://github.com/inferno-framework/inferno-core/issues/754) to support type slice discriminator.path if populated.

PR #286, opened 2026-01-30
Fix type_slices method using correct elements variable

yunwwang:
# Summary

This PR fixes Issue #283 

The first issue is that the `elements` variable used in current version does not exist at all. This PR changes it to the correct instance variable `profile_elements`.

This PR also saves the `type_path` into the metadata if it has value.

Since US Core does not have type slice with path, so this code section has not been touched by US Core generator.

# Testing Guidance

Run `us_core:generate`, there SHALL NOT be any change to the generated us core tests.

### DIFF
--- a/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
+++ b/lib/us_core_test_kit/generator/must_support_metadata_extractor.rb
@@ -84,21 +84,25 @@ module USCoreTestKit
           type_path = '' if type_path == '$this'
           type_element =
             if type_path.present?
-              elements.find { |element| element.id == "#{current_element.id}.#{type_path}" }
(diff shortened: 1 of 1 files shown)


karlnaden (changes requested):
Not convinced yet that this non-functional for this test kit change is worth making since we're essentially maintaining code without a clear use except to let people copy it into their new test kits.

If you still think this important to add, provide a positive test for this, like a modified US Core profile that will exercise this functionality and generate the right metadata (presumably the running test will use the other fix to fhir_resource_navigatioin).

karlnaden (approved):
tested against the PAS v2.2.0 version which uses this. metadata generation looks good.

PR #290, opened 2026-05-08
gem dependency updates and addressing vulnerabilities + non-FHIR:reference filter added

FlexonyoPizza:
- Filler gem path to inferno-core, tls-test-kit, and smart-app-test-kit in order to bundle install and update gemlock locally. Will remove those gem path and reference the actual inferno-core, tls-test-kit, and smart-app-test-kit versions once those gems are released.

- Added a filter in lib/reference_resolution_test.rb that filters out non-FHIR:reference variants passed to 
"def unresolved_references(resources = [])"; 

Some suite tests (in particular medication[x] tests) in the g10 test kit wouldn't pass because 'CodeableConcept' variants were being passed to that function in US-core that only accepts FHIR:reference variants.


lib/us_core_test_kit/reference_resolution_test.rb
```diff
@@ -84,7 +84,7 @@ def unresolved_references(resources = [])
           found_one_reference = false
 
           resolve_one_reference = resources.any? do |resource|
-            value_found = resolve_path(resource, path)
+            value_found = resolve_path(resource, path).select { |v| v.is_a?(FHIR::Reference) }
```
karlnaden, line 87: This looks like it could be the right fix. However, we need at least a spec test case in the `reference_test_spec.rb` and ideally an example execution in the UI that demonstrates that it is resolved. Looks like I did make the same change in a branch I'm working on here, but haven't gotten around to a spec test yet. So I think it is right and should be relatively easy to get a test that proves it.

karlnaden (changes requested):

PR #291, opened 2026-05-13
Gem updates and FHIR:reference guard rail that filters out codeableCo…

FlexonyoPizza:
**Gem Updates in gemspec:**
- 'inferno_core', '~> 1.2', '>= 1.2.2'
- 'smart_app_launch_test_kit',  '~> 1.0', '>= 1.0.1'
- 'tls_test_kit', '~> 1.0', '>= 1.0.2'

**lib/us_core_test_kit/reference_resolution_test.rb line 87:** 
- A guard rail implemented that filters out non-FHIR:reference variants being passed

**spec/us_core/reference_test_spec.rb — two new test groups added:** 
- #unresolved_references with choice type fields (unit test)
Tests the guard rail added at line 87 of reference_resolution_test.rb:
value_found = resolve_path(resource, path).select { |v| v.is_a?(FHIR::Reference) }
Verifies that when resolve_path returns a FHIR::CodeableConcept for a choice-type path (e.g. medication[x]), it is filtered out before any .reference call is made — preventing a NoMethodError crash. Also confirms the path is not counted as unresolved, since no actual reference was found.
- MedicationRequest reference resolution when medication[x] is a CodeableConcept (integration test)
Uses the real USCoreV311::MedicationRequestReferenceResolutionTest class against a stubbed FHIR server returning a MedicationRequest with medicationCodeableConcept. Confirms two things:
1.	No NoMethodError is raised when the guard encounters a CodeableConcept
2.	The test result is pass — the CodeableConcept is silently skipped and the remaining must-support references (e.g. subject) are resolved normally


Gemfile.lock
```diff
@@ -343,5 +376,150 @@ DEPENDENCIES
...
+  unicode-display_width (2.6.0) sha256=12279874bba6d5e4d2728cef814b19197dbb10d7a7837a869bab65da943b7f5a
+  unicode_utils (1.4.0) sha256=b922d0cf2313b6b7136ada6645ce7154ffc86418ca07d53b058efe9eb72f2a40
+  us_core_test_kit (1.1.1)
+  version_gem (1.1.9) sha256=0c1a0962ae543c84a00889bb018d9f14d8f8af6029d26b295d98774e3d2eb9a4
+  webmock (3.26.2) sha256=774556f2ea6371846cca68c01769b2eac0d134492d21f6d0ab5dd643965a4c90
+  zeitwerk (2.7.5) sha256=d8da92128c09ea6ec62c949011b00ed4a20242b255293dd66bf41545398f73dd
+
 BUNDLED WITH
-   2.5.22
+  4.0.7
```
karlnaden: I don't think we want to update the bundler version. Can you:
1. revert Gemfile.lock to its state before your pr
2. regenerate it using `bundle _2.5.22_ update`

karlnaden (changes requested):

karlnaden (approved):

PR #295, opened 2026-07-13
Release v1.1.5 and Client proxying fix

karlnaden:
# Summary

Proxying of request to the client suite FHIR endpoints to the reference server was failing in deployed environments due to a combination of json gem inconsistencies and a varied object structure introduced by the url replacement method. Now, that method keeps the FHIR object structure consistent to resolve the issue.

# Testing Guidance

Deploy to QA and confirm that running the client vs server suite works correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.1.5
+
+* Fix a bug in the client tests proxying.
+
 # 1.1.4
 
 * Adjust v5.0.1 validator message filters for v5.0.1 based on validator version v1.0.79.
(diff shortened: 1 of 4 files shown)


arscan:
This appears to fix the issue in the US Quality Core test kit.  My best guess is that this was a dependency update issue within inferno_core 1.4.0 (perhaps the oj bump?), which this seems to fix?

Note that there are a couple of other test kits with the same issue: https://github.com/search?q=org%3Ainferno-framework+%22def+replace_bundle_urls%22&type=code

ljtucker (approved):

